PR #1, opened 2022-12-08
Useful commands

mcoffey-mt:
Added a section to the readme which contains some commands that people may find useful when working with the kubernetes infrastructure.


mcoffey-mt:
Safe changes to merge. Merging

PR #6, opened 2022-12-15
404 exception

emileswarts:
Throw an exception and return HTTP status code 404 to the client


chubberlisk (approved):
:shipit:

src/test/kotlin/uk/gov/justice/digital/hmpps/hmppsintegrationapi/controllers/PersonControllerTest.kt
```diff
@@ -28,11 +30,21 @@ internal class PersonControllerTest(
...
 
       result.response.status.shouldBe(200)
     }
 
+    it("responds with a 404 NOT FOUND status") {
+      val id_that_does_not_exist = 777
+
+      mockMvc.perform(get("/persons/$id_that_does_not_exist"))
+        .andDo(print())
+        .andExpect(status().isNotFound)
```
chubberlisk, line 45: Could we use the Kotest expectation instead e.g. `result.response.status.shouldBe(404)`?

emileswarts: Ah yes, let me fix

PR #9, opened 2022-12-16
Makefile

emileswarts:
This will allow running tests with one command, shell scripts act as
placeholders for any additional functionality


.circleci/config.yml
```diff
@@ -58,14 +58,12 @@ jobs:
       - run:
           name: Build application Docker image
           command: |
-            docker build -t hmpps-integration-api .
+            make build
```
chubberlisk, line 61: We can also replace the `./gradlew test` within the `unit-tests` job with `make unit-tests`. I guess we could also have one for linting?

chubberlisk (approved):
🚢  - just one smol comment!

PR #12, opened 2022-12-19
Update nomis gateway

mcoffey-mt:
- Additional makefile commands
- ID now requires string instead of int
- New mock server for nomis
- Implemented nomis gateway tests


chubberlisk (approved):
:shipit:

PR #14, opened 2022-12-20
Add try-catch for authentication in NOMIS gateway

chubberlisk:
This is to prevent the application from not starting. In a follow-up PR, we'll look to improve how we handle when an upstream API is not available.


mcoffey-mt (approved):
Looks good!

PR #16, opened 2022-12-22
Add context diagram

chubberlisk:

- Add a basic [context diagram](https://c4model.com/#SystemContextDiagram) to provide a high-level diagram of the systems in play using [PlantUML](https://crashedmind.github.io/PlantUMLHitchhikersGuide/C4/C4Stdlib.html).
- Add context diagram to the about section in README.

## Guidance to review

- The diagram isn't meant to be perfect but something that we can iterative and build upon e.g. it's probably missing a user type connected to MAPPS and relationship lines could be more descriptive.
- Instructions on how to update the diagram will be added in a follow-up PR.
- To preview the README, go to https://github.com/ministryofjustice/hmpps-integration-api/tree/add-context-diagram#about-this-project instead as in the PR it will look broken because of the relative link to the image.


chubberlisk:
> Love it! Merge away!
> 
> Perhaps something to consider for future development on this. Would it be worth adding an additional container alongside mapps which demonstrates another generic consumer which might use ours? I'm very keen on the idea of making it clear that what we're making isn't exclusively for MAPPS.

Yuppppp, really nice idea - we should do definitely do that!

PR #18, opened 2022-12-23
Fix dependency vulnerabilities

chubberlisk:
- Fix dependency vulnerabilities by upgrading `uk.gov.justice.hmpps.gradle-spring-boot` from 4.5.7 to 4.7.4 (latest). 
- Temporarily remove Trivy scan as the orb job fails because it assumes that the Docker image for the application will be pulled from quay.io and doesn't allow us to provide login details. This will be picked up after Christmas.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -112,11 +112,6 @@ workflows:
       - hmpps/gradle_owasp_dependency_check:
           context:
             - hmpps-common-vars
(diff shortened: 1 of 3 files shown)


mcoffey-mt (approved):
Exactly the same as the changes I had intended in my own branch but you beat me to the punch :) Nicely done!

PR #22, opened 2023-01-04
Move test application config file to config directory

emileswarts:
This configuration file will be picked up automatically and saves having to explicitly reference the active profile in tests.


chubberlisk (approved):
:shipit:

PR #24, opened 2023-01-04
Add aliases to get person endpoint from NOMIS

chubberlisk:
## Context

At the moment, our get person endpoint just returns the first name and last name.

- Add aliases found in NOMIS as part of the `GET /api/offenders/{id}` to the response of our `GET /persons/{id}` endpoint e.g.

```json
{
  "firstName": "DANIEL",
  "lastName": "SMELLEY",
  "aliases": [
    {
      "firstName": "DANNY",
      "lastName": "SMILEY"
    }
  ]
}
```

## Guidance to review

For running things locally:

- http://localhost:8080/persons/A1234AL is an example of a person in NOMIS that has an alias.
- http://localhost:8080/persons/A1234AA is an example of a person in NOMIS that doesn't have an alias.

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsintegrationapi/models/Person.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsintegrationapi/models/Person.kt
@@ -21,6 +21,17 @@ data class Person(
 )
 
 data class Alias(
+  @field:Schema(
+    description = "First name",
+    example = "John",
+    type = "string",
+  )
   val firstName: String,
+
+  @field:Schema(
+    description = "Last name",
+    example = "Marston",
+    type = "string",
+  )
   val lastName: String
 )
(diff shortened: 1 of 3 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsintegrationapi/models/Person.kt
```diff
@@ -15,5 +15,23 @@ data class Person(
     example = "Morgan",
     type = "string",
   )
+  val lastName: String,
+
+  val aliases: List<Alias> = listOf()
+)
+
+data class Alias(
```
emileswarts, line 23: Would it be easier to find this Alias model if it lived in a dedicated file?

emileswarts (approved):

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsintegrationapi/models/Person.kt
chubberlisk: Yuppppp, agreed!

PR #28, opened 2023-01-10
Use native docker-compose health checks

emileswarts:
Using this built-in mechanism allows us to delete the custom shell script that used to do the blocking/waiting.


mcoffey-mt (approved):
Very nice! Looks like the right way to do it instead of the shell script variant.

PR #42, opened 2023-02-02
Refactor auth username and password configuration

emileswarts:
No need to specify the auth credentials for each API, we can use the same credentials. Simplify the configuration by promoting this to the root and removing duplicates.


chubberlisk (commented):
Some thoughts:

1. I think that it's still appropriate to have the `hmpps-auth` key under the `services` key as it's an external service like Prison API and Prisoner Offender Search. What do you think?
2. Also, I think we can set the username and password used in `application-local-docker.yml` and `application-local.yml` to the ones used in `application.yml` so we default to that as they're the same. That way we can remove the `hmpps-auth` key completely from them.
3. Do we need to update `src/test/resources/config/application.yml`? Maybe remove some `hmpps-auth` keys?

chubberlisk (approved):
Very nice! 💯

PR #43, opened 2023-02-03
Move username / password values to HmppsAuth gateway

emileswarts:
This is the only class that needs knowledge of these values.


chubberlisk (approved):
Clean!

PR #46, opened 2023-02-07
HIA-104 Network Diagram (Container)

mcoffey-mt:
I'm not 100% happy with the lines not being straight down from Route 53 to APIGateway, but it's likely to change when we decide to add the private link elements.


mcoffey-mt:
Perhaps we can investigate .SVG formats for outputs, as the links in this diagram would then work.

chubberlisk (approved):

mcoffey-mt:
Regarding my comment about .SVG diagrams. We now have a [ticket](https://dsdmoj.atlassian.net/browse/HIA-128) for this

PR #50, opened 2023-02-09
HIA-109 Introduce DTO for Prison API and Prisoner Offender Search

mcoffey-mt:


src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsintegrationapi/models/nomis/OffenderAlias.kt
```diff
@@ -0,0 +1,10 @@
+package uk.gov.justice.digital.hmpps.hmppsintegrationapi.models.nomis
+
+import java.time.LocalDate
+
+data class OffenderAlias(
```
chubberlisk, line 5: Tiny thing but I would call this `Alias` instead as that's the name used in the Prison API.

image

chubberlisk (approved):
Noiceeeeee - just one smol suggestion!

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsintegrationapi/models/nomis/OffenderAlias.kt
mcoffey-mt: Good spot. All for consistent naming of how they have things on their end

PR #59, opened 2023-02-16
Refactor smoke tests to use Kotlin instead of bash

chubberlisk:
## Context

To create our smoke tests, we used bash as a super lightweight solution. However, soon we'll find ourselves writing our own testing framework if we continue this path.

- Remove our bash script for smoke tests and instead use Kotlin and Kotest. This way it's easier to read and we don't have to write bash to output success or failure ourselves.
- Update our CircleCI pipeline to use a machine executor so we can access our containers easily. (Using a Docker executor means we cannot access from our job the remote Docker environment).


emileswarts (approved):

build.gradle.kts
```diff
@@ -25,14 +25,27 @@ java {
...
+  register<Test>("unitTest") {
+    useJUnitPlatform()
+
+    filter {
+      excludeTestsMatching("uk.gov.justice.digital.hmpps.hmppsintegrationapi.smoke*")
+    }
+  }
+
+  register<Test>("smokeTest") {
+    useJUnitPlatform()
```
emileswarts, line 43: Is `useJUnitPlatform()` still required?

chubberlisk: Good spot, forgot to check - nope it isn't!

PR #62, opened 2023-02-20
Added named arguments to DTOs

mcoffey-mt:
Our data transfer objects will soon become quite large when mapping the upstream system's properties to our own. This will result in much larger constructors for our domain models. I've added named arguments to each of the "toX" methods in the data transfer objects so it's crystal clear which fields are being mapped to where. This will also help avoid any fields accidentally being mapped to the wrong field in the future when we have more fields due to the old method being dependent on argument position.


chubberlisk (approved):
Love it! 💯

PR #64, opened 2023-02-21
Added ADR 0004

mcoffey-mt:


mcoffey-mt:
Ignore the schema file that was brought in, it's been deleted in main anyway.

mcoffey-mt:
Once approved, I'll update the product page link to this.

docs/adr/0004-lists-returned-with-root-json-object
```diff
@@ -0,0 +1,55 @@
+# 0004 - Lists returned with root JSON object
```
chubberlisk, line 1: I think this could be clearer about the decision. Maybe "Always return a JSON object for JSON responses" or "Never return a list at the root of a JSON response" or similar.

docs/adr/0004-lists-returned-with-root-json-object
```diff
@@ -0,0 +1,55 @@
...
+  },
+  {
+    "firstName" : "Sue"
+    "lastName" : "Storm"
+  }
+]
+```
+
+_Wrapped in a JSON Object_
+```
```
chubberlisk, line 28:
```suggestion
```json
```

chubberlisk (approved):
Nice! FYI, this file is missing the `.md` file extension.

docs/adr/0004-lists-returned-with-root-json-object
mcoffey-mt: Agreed, I was struggling naming this. I'm happy with "Always return a JSON object for JSON responses"

PR #67, opened 2023-02-22
Add and update OpenAPI specification from our API docs repository

chubberlisk:
## Context

Initially we added our OpenAPI specification file in our API docs (https://github.com/ministryofjustice/hmpps-integration-api-docs/pull/11), however we decided it's better to have it living next to our code so whenever we make changes to our API, it's easy to update and becomes part of our process.

- Add our OpenAPI specification file from our API docs.
- Update the file with latest changes to our API, including adding our endpoints for images.

### DIFF
--- a/openapi.yml
+++ b/openapi.yml
@@ -1,0 +1,76 @@
+openapi: 3.0.0
+info:
+  title: HMPPS Integration API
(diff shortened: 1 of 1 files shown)


mcoffey-mt (approved):
Looks great, it'd be nice if some of our examples used string literals (if this is possible in .yml) but aside from that I like that this is closer to our code base.

PR #69, opened 2023-02-23
Add basic endpoint for getting addresses for a person

chubberlisk:

- Add a `GET /persons/{id}/addresses` that merges addresses found in NOMIS via the Prison API and Probation Offender Search. However, as a first thin slice, we only return the postcode e.g.

```json
{
  "addresses": [
    {
      "postcode": "N12 9UI"
    },
    {
      "postcode": "LI1 5TH"
    }
  ]
}
```

image

## Next steps

- Add more properties to addresses returned.

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsintegrationapi/gateways/NomisGateway.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsintegrationapi/gateways/NomisGateway.kt
@@ -9,10 +9,12 @@ import org.springframework.stereotype.Component
 import org.springframework.web.reactive.function.client.WebClient
 import org.springframework.web.reactive.function.client.WebClientResponseException
 import uk.gov.justice.digital.hmpps.hmppsintegrationapi.exception.EntityNotFoundException
+import uk.gov.justice.digital.hmpps.hmppsintegrationapi.models.Address
 import uk.gov.justice.digital.hmpps.hmppsintegrationapi.models.ImageMetadata
 import uk.gov.justice.digital.hmpps.hmppsintegrationapi.models.Person
 import uk.gov.justice.digital.hmpps.hmppsintegrationapi.models.nomis.ImageDetail
 import uk.gov.justice.digital.hmpps.hmppsintegrationapi.models.nomis.Offender
+import uk.gov.justice.digital.hmpps.hmppsintegrationapi.models.nomis.Address as AddressFromNomis
 
 @Component
 class NomisGateway(@Value("\${services.prison-api.base-url}") baseUrl: String) {
@@ -73,6 +75,24 @@ class NomisGateway(@Value("\${services.prison-api.base-url}") baseUrl: String) {
         .block()!!
     } catch (exception: WebClientResponseException.NotFound) {
       throw EntityNotFoundException("Could not find image with id: $id")
+    }
+  }
+
+  fun getAddressesForPerson(id: String): List<Address> {
+    val token = hmppsAuthGateway.getClientToken("NOMIS")
+
+    return try {
+      webClient
+        .get()
+        .uri("/api/offenders/$id/addresses")
+        .header("Authorization", "Bearer $token")
+        .retrieve()
+        .bodyToFlux(AddressFromNomis::class.java)
+        .map { addressFromNomis -> addressFromNomis.toAddress() }
+        .collectList()
+        .block() as List<Address>
+    } catch (exception: WebClientResponseException.NotFound) {
+      throw EntityNotFoundException("Could not find person with id: $id")
     }
   }
 }
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsintegrationapi/models/nomis/Address.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsintegrationapi/models/nomis/Address.kt
@@ -1,0 +1,9 @@
+package uk.gov.justice.digital.hmpps.hmppsintegrationapi.models.nomis
+
+import uk.gov.justice.digital.hmpps.hmppsintegrationapi.models.Address
+
+data class Address(
+  val postalCode: String
+) {
+  fun toAddress(): Address = Address(postcode = this.postalCode)
+}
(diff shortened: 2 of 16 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsintegrationapi/gateways/NomisGateway.kt
```diff
@@ -9,10 +9,12 @@ import org.springframework.stereotype.Component
 import org.springframework.web.reactive.function.client.WebClient
 import org.springframework.web.reactive.function.client.WebClientResponseException
 import uk.gov.justice.digital.hmpps.hmppsintegrationapi.exception.EntityNotFoundException
+import uk.gov.justice.digital.hmpps.hmppsintegrationapi.models.Address
 import uk.gov.justice.digital.hmpps.hmppsintegrationapi.models.ImageMetadata
 import uk.gov.justice.digital.hmpps.hmppsintegrationapi.models.Person
 import uk.gov.justice.digital.hmpps.hmppsintegrationapi.models.nomis.ImageDetail
 import uk.gov.justice.digital.hmpps.hmppsintegrationapi.models.nomis.Offender
+import uk.gov.justice.digital.hmpps.hmppsintegrationapi.models.nomis.Address as AddressFromNomis
```
mcoffey-mt, line 17: This is really useful, we should look at other opportunities like this to be clear as to what the object is.

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsintegrationapi/models/nomis/Address.kt
```diff
@@ -0,0 +1,9 @@
+package uk.gov.justice.digital.hmpps.hmppsintegrationapi.models.nomis
+
+import uk.gov.justice.digital.hmpps.hmppsintegrationapi.models.Address
+
+data class Address(
+  val postalCode: String
+) {
+  fun toAddress(): Address = Address(postcode = this.postalCode)
```
mcoffey-mt, line 8: You shouldn't need to explicitly specify the type here. Although we're doing it in other cases so I wouldn't bother changing it unless we do it for all, just a thought for refactoring in the future.

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsintegrationapi/models/probationoffendersearch/Address.kt
```diff
@@ -0,0 +1,9 @@
+package uk.gov.justice.digital.hmpps.hmppsintegrationapi.models.probationoffendersearch
+
+import uk.gov.justice.digital.hmpps.hmppsintegrationapi.models.Address
+
+data class Address(
+  val postcode: String
+) {
+  fun toAddress(): Address = Address(postcode = this.postcode)
```
mcoffey-mt, line 8: You shouldn't need to explicitly specify the type here on toAddress(). Although we're doing it in other cases so I wouldn't bother changing it unless we do it for all, just a thought for refactoring in the future.

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsintegrationapi/services/GetAddressesForPersonService.kt
```diff
@@ -0,0 +1,20 @@
...
+@Service
+class GetAddressesForPersonService(
+  @Autowired val probationOffenderSearchGateway: ProbationOffenderSearchGateway,
+  @Autowired val nomisGateway: NomisGateway
+) {
+  fun execute(id: String): List<Address> {
+    val addressesFromProbationOffenderSearch = probationOffenderSearchGateway.getAddressesForPerson(id)
+    val addressesFromNomis = nomisGateway.getAddressesForPerson(id)
+
+    return addressesFromProbationOffenderSearch + addressesFromNomis
```
mcoffey-mt, line 18: `return addressesFromProbationOffenderSearch + addressesFromNomis`

We did the exact same thing in the search endpoint for this. I feel like this could be an area of optimisation in the future when we need to pull details from more than just a couple of systems. Again, just a side-note. No need for action :)

src/test/kotlin/uk/gov/justice/digital/hmpps/hmppsintegrationapi/smoke/ImageSmokeTest.kt
```diff
@@ -13,7 +13,9 @@ class ImageSmokeTest : DescribeSpec({
   val httpClient = HttpClient.newBuilder().build()
   val httpRequest = HttpRequest.newBuilder()
 
-  it("returns an image from NOMIS") {
+  // This test is skipped because the Prison API's endpoint for getting image data/content
+  // has changed and Prism is unable to return a valid response.
+  xit("returns an image from NOMIS") {
```
mcoffey-mt, line 18: Good spot here, is there a Jira ticket to re-add this when it's been fixed?

mcoffey-mt (approved):
Great work. I love the added detail of the diagram in the PR. <insert ship-it gif>

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsintegrationapi/gateways/NomisGateway.kt
chubberlisk: Yeahhhh, it's nice as it means we can still call the models exactly what they are in the external API but when there are ambiguity we can do dis.

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsintegrationapi/models/nomis/Address.kt
chubberlisk: Here it's super clear what the return type is so I've removed it, thanks for the suggestion!

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsintegrationapi/models/probationoffendersearch/Address.kt
chubberlisk: Here it's super clear what the return type is so I've removed it, thanks for the suggestion!

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsintegrationapi/services/GetAddressesForPersonService.kt
chubberlisk: Agreedaroo, one day I can imagine we'll want to sort them or something but for now like you said I think we're good.

src/test/kotlin/uk/gov/justice/digital/hmpps/hmppsintegrationapi/smoke/ImageSmokeTest.kt
chubberlisk: https://dsdmoj.atlassian.net/browse/HIA-160

PR #74, opened 2023-02-28
Add basic endpoint to search person(s) on first and last name

mcoffey-mt:
### Changes proposed in this PR
Add an endpoint to `/persons` which allows the consumer to search on both First name **AND** Last name.

The work in this PR should be considered a 'thin slice' which hits the upstream system's search endpoints respectively and passes the results forward.

<<diagram incoming by @AprilMD >>

### Next Steps
Remove the requirement for either first and last name so that searching can be done on one or the other.

### Notes
1. If no results are found, the API will return an empty list of persons.
2. Under the hood, Prisoner Offender Search requires the presence of a 'includeAliases' property within the JSON body. This has currently been hard-coded in the Gateway,
3. Under the hood, Probation Offender Search requires the presence of a 'valid' property within the JSON body. This has currently been hard-coded in the Gateway,


src/test/kotlin/uk/gov/justice/digital/hmpps/hmppsintegrationapi/gateways/probationoffendersearch/ProbationOffenderSearchGatewayTest.kt
```diff
@@ -26,13 +31,57 @@ class ProbationOffenderSearchGatewayTest(
...
+    Mockito.reset(hmppsAuthGateway)
+
+    whenever(hmppsAuthGateway.getClientToken("Probation Offender Search")).thenReturn(HmppsAuthMockServer.TOKEN)
+  }
+
+  afterTest {
+    probationOffenderSearchApiMockServer.stop()
+  }
+
+  describe("#getPerson(s)") {
```
chubberlisk, line 46:
```suggestion
  describe("#getPersons") {
```

Similar to previous comment!

src/test/kotlin/uk/gov/justice/digital/hmpps/hmppsintegrationapi/services/GetPersonServiceTest.kt
```diff
@@ -25,59 +25,107 @@ internal class GetPersonServiceTest(
...
-    whenever(prisonerOffenderSearchGateway.getPerson(id)).thenReturn(personFromPrisonerOffenderSearch)
-    whenever(probationOffenderSearchGateway.getPerson(id)).thenReturn(personFromProbationOffenderSearch)
+      val expectedResult = listOf(
+        Person(firstName, lastName, middleName = "Gary"),
+        Person(firstName, lastName, middleName = "John")
+      )
 
-    val result = getPersonService.execute(id)
+      val result = getPersonService.execute(firstName, lastName)
+      result.shouldBe(expectedResult)
```
chubberlisk, line 68: I think this can be simplified to:

```kotlin
it("returns person(s)") {
  val personsFromPrisonerOffenderSearch = listOf(Person(firstName, lastName, middleName = "Gary"))
  val personsFromProbationOffenderSearch = listOf(Person(firstName, lastName, middleName = "John"))

  whenever(prisonerOffenderSearchGateway.getPersons(firstName, lastName)).thenReturn(personsFromPrisonerOffenderSearch)
  whenever(probationOffenderSearchGateway.getPersons(firstName, lastName)).thenReturn(personsFromProbationOffenderSearch)

  val result = getPersonService.execute(firstName, lastName)

  result.shouldBe(listOf(personsFromPrisonerOffenderSearch, personsFromProbationOffenderSearch))
}
```

src/test/kotlin/uk/gov/justice/digital/hmpps/hmppsintegrationapi/services/GetPersonServiceTest.kt
```diff
@@ -25,59 +25,107 @@ internal class GetPersonServiceTest(
...
-      "prisonerOffenderSearch" to Person("Sally", "Sob"),
-      "probationOffenderSearch" to Person("Molly", "Mob")
-    )
+    it("returns an empty list when no person(s) are found") {
+      whenever(prisonerOffenderSearchGateway.getPersons(firstName, lastName)).thenReturn(emptyList())
+      whenever(probationOffenderSearchGateway.getPersons(firstName, lastName)).thenReturn(emptyList())
 
-    result.shouldBe(expectedResult)
+      val result = getPersonService.execute(firstName, lastName)
+      result.shouldBe(emptyList())
```
chubberlisk, line 76:
```suggestion
      val result = getPersonService.execute(firstName, lastName)
      
      result.shouldBe(emptyList())
```

Sorry, newlines in tests are the simplest thing to help readability of tests imo. It makes it easier for me to spot the arrange, act and assert parts of a test.

src/test/kotlin/uk/gov/justice/digital/hmpps/hmppsintegrationapi/smoke/PersonSmokeTest.kt
```diff
@@ -14,25 +14,53 @@ class PersonSmokeTest : DescribeSpec({
...
+      """.removeWhitespaceAndNewlines()
+    )
   }
 
-  it("returns image metadata for a person") {
-    val response = httpClient.send(
-      httpRequest.uri(URI.create("$baseUrl/persons/$id/images")).build(),
-      HttpResponse.BodyHandlers.ofString()
-    )
+  describe("returns a person from NOMIS, Prisoner Offender Search and Probation Offender Search when given an id") {
```
chubberlisk, line 44: To me this describe is unnecessary but do as you will!

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsintegrationapi/services/GetPersonService.kt
```diff
@@ -13,6 +13,14 @@ class GetPersonService(
   @Autowired val prisonerOffenderSearchGateway: PrisonerOffenderSearchGateway,
   @Autowired val probationOffenderSearchGateway: ProbationOffenderSearchGateway
 ) {
+
+  fun execute(firstName: String, lastName: String): List<Person?> {
```
chubberlisk, line 17: I don't think that this should be here, instead it should be self-contained in its own service called `GetPersonsService` that has this method description, see https://github.com/madetech/clean-architecture/blob/master/use_case.md#use-case (we call them services but same thing).

>In code, the entry point of a Use Case is a class that has one public method.

This should also help remove the need to add extra describes in tests.

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsintegrationapi/controllers/PersonController.kt
```diff
@@ -20,6 +21,14 @@ class PersonController(
   @Autowired val getImageMetadataForPersonService: GetImageMetadataForPersonService,
   @Autowired val getAddressesForPersonService: GetAddressesForPersonService
 ) {
+
+  @GetMapping
+  fun getPerson(@RequestParam firstName: String, @RequestParam lastName: String): Map<String, List<Person?>> {
```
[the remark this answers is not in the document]
mcoffey-mt: Glad you mentioned this. I actually have a commit sat on my PC here which changes exactly, this. You saying the same thing just affirms that its a positive move :)

src/test/kotlin/uk/gov/justice/digital/hmpps/hmppsintegrationapi/gateways/prisoneroffendersearch/stubGetPrisoners.json
```diff
@@ -0,0 +1,268 @@
+{
```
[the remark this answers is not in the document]
mcoffey-mt: Yep, I like that idea. Done!

src/test/kotlin/uk/gov/justice/digital/hmpps/hmppsintegrationapi/smoke/PersonSmokeTest.kt
mcoffey-mt: Yeeah fair comment. Let's bin it, it just adds confusion

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsintegrationapi/services/GetPersonService.kt
mcoffey-mt: Fair enough, the justification was that we're using the same endpoint so we had grouped them together. But let's stick to the MT way, that makes more sense to me :)

mcoffey-mt: After making this change, the benefits to the tests alone are very nice and I'm fully on board!

chubberlisk (approved):
Thanks for addressing my feedback - ship it! :shipit:

PR #75, opened 2023-02-28
Persons OpenAPI specification

mcoffey-mt:
This will likely need to be expanded upon as we develop the endpoint further. Particularly the required'ness of firstName and lastName. But this is a starting point representing the endpoint in its current state.

### DIFF
--- a/openapi.yml
+++ b/openapi.yml
@@ -293,8 +293,8 @@ components:
         moreInfo: null
     NoQueryParametersBadRequestError:
       value:
-        status: 400
-        errorCode: null
-        userMessage: "400 No query parameters error: There must be at least one query parameter"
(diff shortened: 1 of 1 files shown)


openapi.yml
```diff
@@ -7,6 +7,38 @@ info:
...
+          description: The first name of the person
+        - in: query
+          name: lastName
+          schema:
+            type: string
+          required: true
+          description: The last name of the person
+      responses:
+        "200":
+          description: Successfully performed the query on upstream APIs. May result in an empty list if no results were found.
```
chubberlisk, line 30:
```suggestion
          description: Successfully performed the query on upstream APIs. An empty list is returned when no results are found.
```

chubberlisk (approved):
image

PR #77, opened 2023-03-03
Change unit tests to consistently use HttpStatus

AprilMD:
Previously, unit tests have sometimes checked status codes manually (i.e. result.response.status.shouldBe(404)) and other times used import org.springframework.http.HttpStatus. 

This PR makes the code base consistently use HttpStatus.


chubberlisk (approved):
image

PR #79, opened 2023-03-06
Upgrade to Spring Boot 3

emileswarts:
This upgrades the application to use Spring Boot 3.x Upgrade Gradle
Use latest version of wiremock, which is required with the upgrade, sadly kotest hasn't upgraded their implementation. Remove microsoft log appender, which is no longer compatible.


mcoffey-mt:
Before - Get Persons (Search)
image

Before - Get Persons/{id}
image

mcoffey-mt (approved):
LGTM

PR #81, opened 2023-03-08
Removed unused execute method on GetPersonService

mcoffey-mt:
This was simply a duplicate of the method in GetPersonsService which should be used instead


mcoffey-mt:
This smoke test failure is a known issue, it worked locally. It's the same error @emileswarts had yesterday, this probably increases the priority to fix.

emileswarts (approved):

PR #83, opened 2023-03-09
ADR for Data Transfer Object Pattern

mcoffey-mt:
Documents our decision to use the DTO pattern for mapping from upstream system models to ours. 

[Jira Ticket](https://dsdmoj.atlassian.net/browse/HIA-118)

### DIFF
--- a/docs/adr/0005-use-dto-pattern.md
+++ b/docs/adr/0005-use-dto-pattern.md
@@ -1,0 +1,58 @@
+# 0005 - Data Transfer Object Pattern
+
+2023-03-09
+
+## Status
+
+Accepted
+
+## Context
+
+The system must be capable of retrieving data from numerous data sources; To support this, it will need to be able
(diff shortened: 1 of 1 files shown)


docs/adr/0005-use-dto-pattern.md
```diff
@@ -0,0 +1,58 @@
...
+> By encapsulating the serialization like this, the DTOs keep this logic out of the rest of the code and also provide a clear point
+> to change serialization should you wish."
+> _Source_ [Martin Fowler's Webpage](https://martinfowler.com/eaaCatalog/dataTransferObject.html)
+
+### Example
+Consider the following example.
+
+_Upstream system 'a' model_
+```kotlin
+ data class human(val name: String){
```
emileswarts, line 35: Class name convention is capital H

mcoffey-mt: Good spot! Thanks for that, all three references have been updated

PR #89, opened 2023-03-15
Add string extension for decoding URLs

chubberlisk:
Not sure about the naming of the extension as `encodedPncId.decodeUrl()` looks a bit weird... Wonder if `.decodeUrlCharacters()` or `.decodeUrlEncodedCharacters()` is better, would appreciate feedback on this!

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsintegrationapi/controllers/PersonController.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsintegrationapi/controllers/PersonController.kt
@@ -8,7 +8,7 @@ import org.springframework.web.bind.annotation.RequestMapping
 import org.springframework.web.bind.annotation.RequestParam
 import org.springframework.web.bind.annotation.RestController
 import uk.gov.justice.digital.hmpps.hmppsintegrationapi.exception.EntityNotFoundException
-import uk.gov.justice.digital.hmpps.hmppsintegrationapi.extensions.decodeUrl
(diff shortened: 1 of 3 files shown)


mcoffey-mt:
Looking good Ting. I agree with your comment though. Since it's an extension of string it will be available for any string, so I don't think "decodeURL" really fits as it's a string, not necessarily a URL. I like your suggestion of 

`decodeURLCharacters()` 

I think the word 'decode' implies that it's encoded so you should be covered on that basis.

What happens if you pass non-encoded characters into this? Does it just return the string?

It would be nice to give feedback to the consumer that the value wasn't actually encoded and perhaps throw a 400 bad request. Perhaps in the future we could extend string extensions more to have an `isURLEncoded()` method, then in our controller throw a 400 if the URL isn't encoded.

chubberlisk:
> Looking good Ting. I agree with your comment though. Since it's an extension of string it will be available for any string, so I don't think "decodeURL" really fits as it's a string, not necessarily a URL. I like your suggestion of
> 
> `decodeURLCharacters()`
> 
> I think the word 'decode' implies that it's encoded so you should be covered on that basis.
> 
> What happens if you pass non-encoded characters into this? Does it just return the string?
> 
> It would be nice to give feedback to the consumer that the value wasn't actually encoded and perhaps throw a 400 bad request. Perhaps in the future we could extend string extensions more to have an `isURLEncoded()` method, then in our controller throw a 400 if the URL isn't encoded.

Cool, thanks for the feedback! I've renamed the method and added a test to demonstrate that when a string contains no URL encoded characters it just returns the string back.

PR #90, opened 2023-03-16
Upgraded hmpps gradle springboot from 5.1.0 to 5.1.2

mcoffey-mt:
It looks like the git changelog for hmpps gradle spring boot isn’t being updated so I had to do some manual digging to have any confidence that the issue will be fixed by upgrading from 5.1.0 to 5.1.2.

CVE-2020-8908 should have been suppressed 13 days ago.
image

CVE-2022-45688 was suppressed on 1st Feb.
image

What concerns me here is we definitely upgraded prior to this, version 5.1.0 was released on 8th Feb so we should have had at least the second change.

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -1,5 +1,5 @@
 plugins {
-  id("uk.gov.justice.hmpps.gradle-spring-boot") version "5.1.0"
+  id("uk.gov.justice.hmpps.gradle-spring-boot") version "5.1.2"
   kotlin("plugin.spring") version "1.8.0"
 }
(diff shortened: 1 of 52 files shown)


emileswarts (approved):

mcoffey-mt:
It seems this version of hmpps spring boot gradle makes trailing commas required. This does seem to be now recommended practice and isn't without its benefits
image

We decided to make the format changes across our project in this PR and merge it. From this point on whenever we run make lint/format it should complain about missing commas.

PR #91, opened 2023-03-16
Add Sentry

emileswarts:
DSN contained in k8s secrets, only activated if the environment variable SENTRY_DSN is set.


chubberlisk (approved):
🚢

PR #94, opened 2023-03-17
Add ADR for choice to URL-encode path parameters that contain forward slashes

AprilMD:


docs/adr/0006-url-encode-path-parameters.md
```diff
@@ -0,0 +1,51 @@
...
+
+2023-03-16
+
+## Status
+
+Accepted
+
+## Context
+
+Users must be able to search the system for nominals on Police National Computer (PNC) ID. PNC ID is a unique person
```
mcoffey-mt, line 11: > Users must be able to search the system for nominals on Police National Computer (PNC) ID. PNC ID is a unique person

'nominals' is definitely a word I've heard MAPPS use a lot. I'm not sure whether by using it here we'd be gearing our documentation directly at them rather than a generic consumer of our API. I've asked MoJ what their preferred term is for this.

docs/adr/0006-url-encode-path-parameters.md
```diff
@@ -0,0 +1,51 @@
...
+identifier. RESTful API design convention is that path parameters are used to identify a specific resource or resources.
```
mcoffey-mt, line 12:
```suggestion
identifier. RESTful API design convention would typically state that that path parameters are used to identify a specific resource or resources.
```

docs/adr/0006-url-encode-path-parameters.md
mcoffey-mt: Had a chat with Martin, they would prefer us avoiding this term unless it's directly requested. The preferred alternatives are:
- Person
- Person in Prison
- Person on Probation

docs/adr/0006-url-encode-path-parameters.md
```diff
@@ -0,0 +1,51 @@
...
+However, passing PNC IDs into our API as path parameters is problematised by the fact that PNC IDs contain a forward
+slash (e.g., 12/345345B) which is the URO standard path delimiter.
```
mcoffey-mt, line 14:
```suggestion
slash (e.g., 12/345345B) which is the path delimiter.
```

docs/adr/0006-url-encode-path-parameters.md
```diff
@@ -0,0 +1,51 @@
...
+
+We considered two options for dealing with the forward slash in PNC ID:    
+**1. Encode the PNC ID**
+
+- The PNC ID 12/345345B would be passed to the /persons end-point URL-encoded as 12%2F345345B.
+
+Considerations:
+
+- This would result in some added complexity for the consumer, who would have to encode the PNC ID.
```
mcoffey-mt, line 23: We also need to make sure our documentation clearly indicates that it's encoded and it's on them to do the encoding.

docs/adr/0006-url-encode-path-parameters.md
```diff
@@ -0,0 +1,51 @@
...
+
+**2. Use query parameters to pass the PNC ID to the API**
+
+- The request format would become /persons?pnc_id={pncId} rather than /persons/{pncId}.
+
+Considerations:
+
+- Query parameters should be used to sort or filter on resources and so this usage is not best practice.
```
mcoffey-mt, line 31: I'm not sure I agree with this comment, we're using query parameters on our search endpoint to search for people by first and last name. You can definitely use query params to control pagination and sorting for sure, but I don't think its necessarily search on them. Perhaps the take away here is that when searching on a single criteria such as an identifier the preferred method would be to do it via the url path.

docs/adr/0006-url-encode-path-parameters.md
```diff
@@ -0,0 +1,51 @@
...
+- It is best practice to use logical nesting on end-points to show relationships. Using query parameters for PNC ID is
+  likely to cause confusion by detaching it from that with which it is most closely associated. An example is
+  persons/images?pnc_id=id, where PNC ID relates foremost to persons rather than images but the URL distorts this
+  relationship.
+
```
mcoffey-mt, line 36: It might be worth mentioning our images or addresses endpoint which would become nasty. Currently its `/persons/{pncId}/images` 
It would need to be
`/persons/images?pnc_id=12/24112F`

docs/adr/0006-url-encode-path-parameters.md
mcoffey-mt: I see you have this as a consequence, this can be ignored.

mcoffey-mt (approved):
Couple of small changes! Feel free to merge after you've had a look :)

emileswarts (approved):

PR #98, opened 2023-03-21
Add testing strategy diagram

AprilMD:


docs/diagrams/testing-strategy.puml
```diff
@@ -0,0 +1,61 @@
...
+
+System_Ext(consumer, "Consumer Application", "An authenticated consumer of the Integration API.")
+
+Boundary(spring_boot_api, "HMPPS Integration REST API", "Spring Boot Kotlin API"){
+  Container(controller, "Controller", "Spring Boot API Controller", "")
+  Container(service, "Service", "Spring Boot Service Bean", "")
+  Container(gateway, "Gateway", "Spring Boot Component Bean", "")
+}
+
+Boundary(unit_test_boundary, "Unit Tests"){
```
mcoffey-mt, line 28: Is it worth including that these are in written within Kotlin as a tag underneath?

docs/diagrams/testing-strategy.svg
mcoffey-mt, line 1: I'm wondering whether we should illustrate via a boundary that the Unit Tests live within the Kotlin application. I guess it depends who the audience of this diagram is? If it's an engineer then this would be useful information, if it's a consumer then it probably wouldn't.

emileswarts (approved):

mcoffey-mt (approved):
Very nice!

docs/diagrams/testing-strategy.svg
AprilMD: We've tried to clarify the relationships of the components of the tests to the external systems, such that we think the external systems now feel relevant to the diagram. Interested to get your view?

mcoffey-mt (approved):
Very nice. You've taken something that was already really good and turned it into something perfect! I think this more clearly demonstrates that our Mock Servers are pretending to be the Upstream Systems and that the smoke tests are pretending to be Consumer Applications.

I may need to get some tips on formatting plantuml from you two because this is fantastic!

PR #102, opened 2023-03-22
Renamed deployment diagram to system landscape

mcoffey-mt:
I had a battle with formatting on this one today and it's really tricky to get the lines to behave in a consistent way. It won the battle.

1. A comment was made about how the Route53 actually provides the IP address for the systems to go to, I decided not to illustrate this on the diagram as it's a level of detail I'm not sure we need at this point. I agree that it makes it look like the diagram routes the request through. But as far as the consumer is concerned that's the behaviour they get as they make a request on a URL, they don't necessarily know how Route53 is handling that.
2. Added missing mTLS exchange between 'any other organisation' and API Gateway
3. We now have a logging & metrics boundary, this just shows what other services are at play. 
4. Another comment was made about whether we should display upstream systems in here. I decided not to include them because they are within the Integration API, the container.svg diagram illustrates them.


docs/diagrams/system-landscape.puml
```diff
@@ -2,7 +2,8 @@
 
    !include https://raw.githubusercontent.com/plantuml-stdlib/C4-PlantUML/master/C4_Context.puml
 
-   title HMPPS Integration API - Deployment Diagram to illustrate connectivity
+   title HMPPS Integration API - System Landscape
```
chubberlisk, line 5: I just realised that in our other diagrams, we have it written like `[Context] HMPPS Integration API`.

```suggestion
   title [System Landscape] HMPPS Integration API
```

docs/diagrams/system-landscape.puml
```diff
@@ -15,24 +16,37 @@
    }
```
chubberlisk, line 16: I know the MAPPS and other organisation existed before but I'm wondering if we should replace these with our "Consumer Application" container for consistency?

docs/diagrams/system-landscape.puml
```diff
@@ -15,24 +16,37 @@
...
+         Rel(apigateway, s3, "Checks certificates against")
+      }
+
+      Boundary(logging, "Logging & Metrics"){
+         System(kibana, "Kibana", "Performance metrics.")
+         System(insights, "Azure Insights", "Usage metrics.")
+         System(sentry, "Sentry", "Error logging.")
+      }
+
       Boundary(cloudplatform, "Cloud Platform", "Elastic Kubernetes Service"){
```
chubberlisk, line 34: I have a fun question, should Cloud Platform live within AWS? I'm not sure of my opinion if I'm honest!

chubberlisk (approved):
Two questions but happy to go with whatever you think so approved!

chubberlisk:
Also, could you update https://github.com/ministryofjustice/hmpps-integration-api/tree/main/docs/diagrams with this diagram please? We added this yesterday so you may have to rebase your branch.

docs/diagrams/system-landscape.puml
AprilMD: (+1 for yes)

mcoffey-mt:
We've done this in another PR

PR #103, opened 2023-03-23
Set Sentry environment

emileswarts:
Currently all environments are seen as production, set the correct environments explicitly with an environment variable.


chubberlisk (approved):
:shipit:

PR #108, opened 2023-03-24
Update API endpoints to return PNC ID for a person

chubberlisk:
- Add PNC ID to our Person model and update DTOs for Prisoner Offender Search and Probation Offender Search.
- Update OpenAPI spec to document PNC ID in Person model.


mcoffey-mt (approved):
Looks good. Merge away. I had a moment of confusion when looking at the PersonSmokeTest.kt because of the types in probationOffenderSearch. But realised this is because they don't have any examples in their spec so prism takes the data types.

PR #110, opened 2023-03-27
Update docker compose healthcheck

emileswarts:
Send the required version header
We think this is an OpenAPI issue where an optional header with a default is still required

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -38,7 +38,7 @@ services:
       dockerfile: Dockerfile.setup-prison-api
     container_name: prison-api
     healthcheck:
-      test: 'wget --header="Authorization: Bearer abc" http://127.0.0.1:4010/api/offenders/A1234AL -O /dev/null'
+      test: 'wget --header="Authorization: Bearer abc" --header="version: 1.0" http://127.0.0.1:4010/api/offenders/A1234AL -O /dev/null'
     ports:
       - '4030:4010'
 


src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsintegrationapi/extensions/WebClientWrapper.kt
```diff
@@ -12,10 +12,10 @@ class WebClientWrapper(
     .baseUrl(baseUrl)
     .build()
 
-  inline fun <reified T> request(method: HttpMethod, uri: String, token: String, requestBody: Map<String, Any?>? = null): T {
+  inline fun <reified T> request(method: HttpMethod, uri: String, headers: Map<String, String>, requestBody: Map<String, Any?>? = null): T {
     val responseBodySpec = client.method(method)
       .uri(uri)
-      .header("Authorization", "Bearer $token")
+      .headers { header -> headers.forEach { h -> header.set(h.key, h.value) } }
```
chubberlisk, line 18: Maybe we name the parameter `requestHeader` and then we don't need to use just `h`?

docker-compose.yml
```diff
@@ -38,7 +38,7 @@ services:
       dockerfile: Dockerfile.setup-prison-api
     container_name: prison-api
     healthcheck:
-      test: 'wget --header="Authorization: Bearer abc" http://127.0.0.1:4010/api/offenders/A1234AL -O /dev/null'
+      test: 'wget --header="Authorization: Bearer abc" --header="version: 1.0" http://127.0.0.1:4010/api/offenders/A1234AL -O /dev/null'
```
chubberlisk, line 41: I think we can remove dis now.

chubberlisk (approved):

PR #112, opened 2023-03-29
Document backups and persistence

emileswarts:


README.md
```diff
@@ -86,6 +87,15 @@ Nodes are spread across multiple (up to 3) Availability Zones.
...
 Kubernetes with auto heal and prevent any corrupt nodes from replacing existing healthy ones.
 
+## Backups
+
+This is a transient API with no database, no data passing through the API is persisted.
+
+Only the following is persisted:
+- Mutual TLS certificate authority - S3 (versioning enabled)
+- API container Images - ECR (versioning enabled)
+- Application logs - AWS hosted ElasticSearch cluster (retention period of 30 days)
```
chubberlisk, line 97: What about Azure Application Insights?

chubberlisk (approved):

README.md
emileswarts: Could become a maintenance burden to list all dependent services. 
There are services we've not touched on like CloudTrail, Prometheus, API Gateway, Sentry .etc
Backups are usually most interesting to the disaster recovery audience, tempted to remove "Application logs" and limit this list to core components?

chubberlisk: Yeahhhh, happy to go with that!

PR #113, opened 2023-03-29
NoSuchElementException searching for person on PNC Id

mcoffey-mt:
When searching for a person, if PrisonerOffenderSearch returned no results, but ProbationOffenderSearch did, we were getting an EmptyList exception. More detail can be found on the exception in Sentry.

```
{
    "status": 500,
    "errorCode": null,
    "userMessage": "Unexpected error: List is empty.",
    "developerMessage": "List is empty.",
    "moreInfo": null
}
```

This is because we were using .first() without first checking the list is empty.

We do check if BOTH results are empty/null and return null if this was the case. But we should have also checked this one explicitly.

I've also added a unit test which helped me test my fix, and will help us prevent regression.

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsintegrationapi/services/GetPersonService.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsintegrationapi/services/GetPersonService.kt
@@ -12,15 +12,17 @@ class GetPersonService(
   @Autowired val probationOffenderSearchGateway: ProbationOffenderSearchGateway,
 ) {
   fun execute(pncId: String): Map<String, Person?>? {
-    val personFromPrisonerOffenderSearch = prisonerOffenderSearchGateway.getPersons(pncId = pncId)
+    val personsFromPrisonerOffenderSearch = prisonerOffenderSearchGateway.getPersons(pncId = pncId)
     val personFromProbationOffenderSearch = probationOffenderSearchGateway.getPerson(pncId)
 
-    if (personFromPrisonerOffenderSearch.isEmpty() && personFromProbationOffenderSearch == null) {
+    if (personsFromPrisonerOffenderSearch.isEmpty() && personFromProbationOffenderSearch == null) {
       return null
     }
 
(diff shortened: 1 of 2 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsintegrationapi/services/GetPersonService.kt
```diff
@@ -12,15 +12,17 @@ class GetPersonService(
...
-    if (personFromPrisonerOffenderSearch.isEmpty() && personFromProbationOffenderSearch == null) {
+    if (personsFromPrisonerOffenderSearch.isEmpty() && personFromProbationOffenderSearch == null) {
       return null
     }
 
+    val personFromPrisonerOffenderSearch = if (personsFromPrisonerOffenderSearch.isEmpty()) null else personsFromPrisonerOffenderSearch.first()
+
     return mapOf(
-      "prisonerOffenderSearch" to personFromPrisonerOffenderSearch.first(),
+      "prisonerOffenderSearch" to personFromPrisonerOffenderSearch,
```
chubberlisk, line 25:
```suggestion
      "prisonerOffenderSearch" to personFromPrisonerOffenderSearch.firstOrNull(),
```

Could we use `firstOrNull()` instead? Looks like it does the same thing as the if statement above.

image

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsintegrationapi/services/GetPersonService.kt
```diff
@@ -12,15 +12,17 @@ class GetPersonService(
   @Autowired val probationOffenderSearchGateway: ProbationOffenderSearchGateway,
 ) {
   fun execute(pncId: String): Map<String, Person?>? {
-    val personFromPrisonerOffenderSearch = prisonerOffenderSearchGateway.getPersons(pncId = pncId)
+    val personsFromPrisonerOffenderSearch = prisonerOffenderSearchGateway.getPersons(pncId = pncId)
```
chubberlisk, line 15: Nice. 😎

emileswarts (approved):

PR #118, opened 2023-04-04
Add explicit platform configuration to docker builds

emileswarts:
Without this, any build from a M1 host machine will result in an error when running the container: "exec format error"


chubberlisk:
Closing as this didn't fix things for us... :(( We'll have to pick this back up another time!

PR #134, opened 2023-04-26
Enable alerts for failed builds and new releases

emileswarts:
These notifications will be sent to our Slack alerts channel


chubberlisk (approved):
:shipit:

PR #135, opened 2023-04-26
Replace default alerts with alerts on failure only

emileswarts:
For now we're only interested in build pipeline failures being reported to Slack.

This should alert on the following:
1. Failed automated tests
2. Failed build/deploy in any of the 3 environments
3. Failed dependency checks
4. Failed security scans


chubberlisk (approved):
Very nice, didn't about this `<<: *slack-fail-post-step` syntax. Ship it! 🚢

PR #158, opened 2023-05-31
Update local deploy script

emileswarts:
Since the secrets have changed, this script needs to updated to reference the correct ECR credentials.


chubberlisk (approved):
:shipit:

PR #161, opened 2023-06-01
Add documentation to monitor all pod logs at once

mcoffey-mt:
Since our pods are load balanced, it's useful to be able to monitor multiple pod's logs at once.

Credit to @emileswarts for sussing out that we can do this via labels.

### DIFF
--- a/docs/useful-commands.md
+++ b/docs/useful-commands.md
@@ -49,6 +49,12 @@ To view logs of a pod:
 kubectl logs <pod-name> -n <namespace>
 ```
 
+To monitor all pod logs in a namespace at once
+```bash
+kubectl get pods --show-labels -n <namespace>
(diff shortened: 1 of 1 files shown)


docs/useful-commands.md
```diff
@@ -49,6 +49,12 @@ To view logs of a pod:
 kubectl logs <pod-name> -n <namespace>
 ```
 
+To monitor all pod logs in a namespace at once
+```bash
+kubectl get pods --show-labels -n <namespace>
+kubectl logs -n <namespace> -l pod-template-hash=<hash> -f
```
chubberlisk, line 55: What's the `<hash>`? Is that what you get from the first command?

chubberlisk (approved):

PR #164, opened 2023-06-02
Refactor get an image to use response wrapper class

chubberlisk:
## Context

In #160, we introduced a wrapper class to be used for returning data from gateways and services.

- Refactor get an image to use the wrapper class for consistency.
- Move throwing `EntityNotFoundException` at the controller level instead.


mcoffey-mt (approved):

mcoffey-mt:
Looks good Ting!

PR #174, opened 2023-06-14
Don't hardcode "England" as country for probation

emileswarts:
This field isn't present in the upstream API, leave it as null


luke-a-williams (commented):
Just leaving this to say I've caught up and seen this content and agree with this decision.

PR #183, opened 2023-06-16
Update heartbeat to retry 5 times in 10 secs intervals before failing

chubberlisk:
## Context

We noticed that when our heartbeat script runs after deployment, there's a possibility that it could poke the previous version of the API i.e. not the latest deployment version as the Kubernetes pods gets switched over.

- Update heartbeat to retries multiple times before declaring a failure. Atm, it's configured to retry 5 times and wait 10 seconds after the next retry.
- Fix `./scripts/heartbeat.sh: line 14: [: =: unary operator expected` error.
- Add clearer logging i.e. add new lines, prefix logs, add emojis (fite me on dis).
- Update heartbeat job names in our pipeline as they previously were the same in the deployment pipeline and the continuous job - now a job in a deployment pipeline is named like `post-deploy-check-heartbeat-for-dev` and the continuous job is `check-heartbeat-for-dev`.

## Screenshots

Before | After
-- | --
image  |  image
image | image

## Next steps

- Output error from curl command so we know what caused a failure.
- Add time to logs.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -174,7 +174,7 @@ workflows:
 
       - heartbeat:
           <<: *slack-fail-post-step
-          name: heartbeat-dev
+          name: post-deploy-check-heartbeat-for-dev
           context:
             - hmpps-common-vars
             - hmpps-integration-api-heartbeat-dev
@@ -188,7 +188,7 @@ workflows:
             - hmpps-common-vars
             - hmpps-integration-api-preprod
           requires:
-            - heartbeat-dev
+            - post-deploy-check-heartbeat-for-dev
           filters:
             branches:
               only:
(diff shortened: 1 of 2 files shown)


mcoffey-mt (approved):
Absolutely love these changes, I will not fight you on the emojis. You've really cleaned this up and made it shine. Very nice Ting!

The only thing I'll mention is status codes, in the ticket we had mentioned that we wanted to include them as an improvement. But I suspect you may be working on this one individually?

chubberlisk:
> Absolutely love these changes, I will not fight you on the emojis. You've really cleaned this up and made it shine. Very nice Ting!
> 
> The only thing I'll mention is status codes, in the ticket we had mentioned that we wanted to include them as an improvement. But I suspect you may be working on this one individually?
> 
> 📦

Thanks Matt! Yeahhhh, adding the failing status code is my next step.

PR #192, opened 2023-06-23
Implement basic offence information endpoint

AprilMD:
Added an initial offences endpoint. Notably:
- Offence histories only currently come from Prison API.
- We've exposed 3 fields only: date, offence code, and description. These will be expanded in line with requirements.

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsintegrationapi/controllers/v1/person/OffencesController.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsintegrationapi/controllers/v1/person/OffencesController.kt
@@ -1,7 +1,11 @@
 package uk.gov.justice.digital.hmpps.hmppsintegrationapi.controllers.v1.person
 
 import org.springframework.beans.factory.annotation.Autowired
-import org.springframework.web.bind.annotation.*
+import org.springframework.web.bind.annotation.GetMapping
+import org.springframework.web.bind.annotation.PathVariable
+import org.springframework.web.bind.annotation.RequestMapping
+import org.springframework.web.bind.annotation.RequestParam
+import org.springframework.web.bind.annotation.RestController
 import uk.gov.justice.digital.hmpps.hmppsintegrationapi.exception.EntityNotFoundException
 import uk.gov.justice.digital.hmpps.hmppsintegrationapi.extensions.decodeUrlCharacters
 import uk.gov.justice.digital.hmpps.hmppsintegrationapi.models.Offence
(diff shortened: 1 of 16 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsintegrationapi/models/Offence.kt
```diff
@@ -0,0 +1,9 @@
+package uk.gov.justice.digital.hmpps.hmppsintegrationapi.models
+
+import java.time.LocalDate
+
+data class Offence(
+  val date: LocalDate?,
+  val code: String?,
+  val description: String?,
+)
```
mcoffey-mt, line 9: I think these can be non-nullable based on the fact that we're only pulling from NOMIS right now, and these fields in NOMIS are not nullable

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsintegrationapi/controllers/v1/person/OffencesController.kt
```diff
@@ -0,0 +1,39 @@
...
+import uk.gov.justice.digital.hmpps.hmppsintegrationapi.models.Offence
+import uk.gov.justice.digital.hmpps.hmppsintegrationapi.models.UpstreamApi
+import uk.gov.justice.digital.hmpps.hmppsintegrationapi.models.UpstreamApiError
+import uk.gov.justice.digital.hmpps.hmppsintegrationapi.services.GetOffencesForPersonService
+import uk.gov.justice.digital.hmpps.hmppsintegrationapi.util.PaginatedResponse
+import uk.gov.justice.digital.hmpps.hmppsintegrationapi.util.paginateWith
+
+@RestController
+@RequestMapping("/v1/persons")
+class OffencesController(
```
chubberlisk, line 20: Really like this approach of putting this under a `controllers/v1/person` package, makes a controller smol!

src/test/kotlin/uk/gov/justice/digital/hmpps/hmppsintegrationapi/gateways/nomis/GetOffencesForPersonTest.kt
```diff
@@ -0,0 +1,106 @@
...
+
+        Mockito.reset(hmppsAuthGateway)
+        whenever(hmppsAuthGateway.getClientToken("NOMIS")).thenReturn(HmppsAuthMockServer.TOKEN)
+      }
+
+      afterTest {
+        nomisApiMockServer.stop()
+      }
+
+      describe("#getOffences") {
```
chubberlisk, line 74: This `describe` is redundant now that each file only tests one gateway method.

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsintegrationapi/models/Offence.kt
emileswarts: Agreed

src/test/kotlin/uk/gov/justice/digital/hmpps/hmppsintegrationapi/models/nomis/OffenceTest.kt
```diff
@@ -0,0 +1,25 @@
+package uk.gov.justice.digital.hmpps.hmppsintegrationapi.models.nomis
+
+import io.kotest.core.spec.style.DescribeSpec
+import io.kotest.matchers.shouldBe
+import java.time.LocalDate
+
+class OffenceTest : DescribeSpec(
```
chubberlisk, line 7: This should be `OffenceHistoryDetailTest` and the file name should be the same.

emileswarts (approved):

chubberlisk (approved):
Very noice work - like the setup with the controller and well done for remembering to update the OpenAPI spec too!

PR #193, opened 2023-06-26
Fix get persons endpoint not returning all Prisoner Offender Search results

chubberlisk:
Prisoner Offender Search's /global-search API endpoint returns a paginated response however we didn't have any logic to handle that. As a result, if the results from it was more than the default page size (10 in this case), then our API never returned them and incorrectly returned the total count between all upstream APIs used for this endpoint.

This fix is a temporary which attempts to get all the results from Prisoner Offender Search by providing a page size of 9999 (an arbitrary number). We have a ticket in the backlog to find a long term solution to dealing with pagination from upstream APIs.


mcoffey-mt (approved):
As nice as the 9999 solution can be! Go for it!

As discussed in slack, we don't love the name of the variable and the scope could be elevated if we have more instances of it. But since it's only used in this scenario as a temporary measure; I think it's fine for now.

PR #195, opened 2023-06-29
Update offence endpoint to include more fields from our golden record

luke-a-williams:


openapi.yml
```diff
@@ -535,19 +535,33 @@ components:
...
+          description: Latest court date associated with the offence
         description:
           type: string
           example: Commit an act / series of acts with intent to pervert the course of public justice
           description: Description associated with the CJS offence code
+        endDate:
+          type: string
+          format: date
+          example: 2018-03-10
+          description: End date of range the offence was believed to have taken place
```
AprilMD, line 555: "End date of range _over which_ the offence was believed to have taken place"

openapi.yml
```diff
@@ -535,19 +535,33 @@ components:
     Offence:
       type: object
       properties:
-        date:
-          type: string
-          format: date
-          example: 1965-12-01
-          description: Date the offence took place or start date if the offence occurred over a period of time
-        code:
+        CJSCode:
```
AprilMD, line 538: personally unsure whether we'd say CJSCode or cjsCode - probably check w someone else

AprilMD (approved):
Left a couple of teenytiny comments but looks good to me

openapi.yml
luke-a-williams: Yes and I completely agree with you - the description that is there is pulled from the grammatically incorrect one from the Prison API

openapi.yml
luke-a-williams: Yes and good spot, that should be a lowercase

PR #197, opened 2023-06-30
Rename prisonerId to nomisNumber

chubberlisk:
This is to make the naming of other identifiers used for a person consistent.


mcoffey-mt (approved):
All good!

PR #204, opened 2023-07-04
Fix empty string for middle names of offenders in Probation

chubberlisk:
When an offender from Probation Offender Search had no middle names, our API would return `""` i.e. an empty string. This fixes this to return null instead.


mcoffey-mt (approved):
Good spot! Approved

PR #214, opened 2023-07-19
initialise schema hierarchy script

luke-a-williams:
- Updated READMEs
- added schema diagram generation
- added schema parent-child data output


scripts/data_analysis/README.md
```diff
@@ -34,8 +35,33 @@ To run this script *from within the tooling folder*, based on your python distri
...
+[generate_schema_diagram.py](generate_schema_diagram.py)
+There are several options for the running of this script:
+- Search for one schema
+- Search for multiple schema
+- No Search option - generate full diagram
+
+To run this script *from within the tooling folder*, based on your python distribution:
+- Search for one schema, where the argument is a schema name in a string format
+```shell
+python3 generate_schema_diagram.py arg1
```
chubberlisk, line 51:
```suggestion
python3 generate_schema_diagram.py arg1
# E.g. python3 generate_schema_diagram.py sentence
```

I like to provide an example for shell command, could we provide one for each of the below commands? E.g. https://github.com/ministryofjustice/hmpps-integration-api/blob/main/docs/useful-commands.md#kubectl

scripts/data_analysis/README.md
```diff
@@ -34,8 +35,33 @@ To run this script *from within the tooling folder*, based on your python distri
...
+>A csv file in the [outputs](outputs) directory containing schemas and fields within them
+
+### Generate a Schema space diagram, output child-parent relations
+[generate_schema_diagram.py](generate_schema_diagram.py)
+There are several options for the running of this script:
+- Search for one schema
+- Search for multiple schema
+- No Search option - generate full diagram
+
+To run this script *from within the tooling folder*, based on your python distribution:
```
chubberlisk, line 48:
```suggestion
To run this script *from within the tooling folder*, based on your Python distribution:
```

scripts/data_analysis/README.md
```diff
@@ -46,4 +72,4 @@ This allows you to edit constants in one location, without having to amend other
 Feel free to explore them for more functionality.
 
 ## Contributing
-Contributions to this tooling section are welcome, as long as they can be executed in python. Autodocumentation is a potential and the desire is to keep this option open as this tooling section expands.
\ No newline at end of file
+Contributions to this tooling section are welcome, as long as they can be executed in python. Autodocumentation is a potential and the desire is to keep this option open as this tooling section expands.
```
chubberlisk, line 75:
```suggestion
Contributions to this tooling section are welcome, as long as they can be executed in Python. Autodocumentation is a potential and the desire is to keep this option open as this tooling section expands.
```

scripts/data_analysis/README.md
```diff
@@ -34,8 +35,33 @@ To run this script *from within the tooling folder*, based on your python distri
 ```shell
 python3 retrieve_schema_fields.py
 ```
-Outputs:
-A csv file in the [outputs](outputs) directory containing
+>Outputs:
+>A csv file in the [outputs](outputs) directory containing schemas and fields within them
```
chubberlisk, line 39:
```suggestion
>Outputs:
> - A csv file in the [outputs](outputs) directory containing schemas and fields within them
```

I know it's only one output for this one but to make it consistent with outputs of other script and visually imply that there could be multiple outputs?

scripts/data_analysis/constants/common.py
```diff
@@ -1,8 +1,18 @@
 import pandas as pd
 import requests
 import yaml
+import os
 
 URL = 'https://api-dev.prison.service.justice.gov.uk/v3/api-docs'
+SCHEMA_FIELD_FILE = "outputs/Schema_field.csv"
+SCHEMA_PARENT_CHILD_FILE = "outputs/Schema_parent_child.csv"
+SCHEMA_DIAGRAM = "outputs/schema_hierachy.dot"
```
chubberlisk, line 9: Should all the naming of the files be consistent? Or is this on purpose to have the CSV files begin with a capital and the .dot file different?

scripts/data_analysis/constants/common.py
```diff
@@ -25,3 +35,47 @@ def extract_data(url=URL):
     else:
         print(url, " responded with ", response.status_code)
         return '{}'
+
+def findParentSchema(response_dict, schema_name):
+    """
+    Search all schemas for any reference to target schema (i.e. find all direct parents) and return results in a dataframe.
```
chubberlisk, line 41: Wonder if `schema_name` should be named `target_schema` to follow the language of the function description? Or maybe in the description just add `schema_name` to link the two?

scripts/data_analysis/constants/common.py
```diff
@@ -25,3 +35,47 @@ def extract_data(url=URL):
...
+    This function assumes the dictionary object passed in is generated from OpenAPI Swagger spec
+
+        Parameters:
+            response_dict (dict): A dictionary object representing the extract from API Docs, such as the output of the extract_data function above
+            schema_name (str): A string of the exact schema name to search for
+
+        Returns:
+            df (pd.DataFrame): a dataframe of schema, field and reference information.
```
chubberlisk, line 49: Personally, not a fan of variables that are acronyms. Would prefer `data_frame` and `data_frames` especially helpful a not very experienced Python person. However I'm also on the fence as I've seen this kind of pattern in Python befoar! 😅

scripts/data_analysis/generate_schema_diagram.py
```diff
@@ -0,0 +1,73 @@
...
+OUTPUT_FILE = common.SCHEMA_PARENT_CHILD_FILE
+DIAGRAM_FILE = common.SCHEMA_DIAGRAM
+
+response_dict = common.extract_data(common.URL)
+
+def main():
+    common.check_directory(DIAGRAM_FILE)
+    data_df = pd.DataFrame()
+
+    if len(sys.argv) == 1: #No arguments passed in - full schema diagram
```
chubberlisk, line 20:
```suggestion
    is_full_schema_diagram = len(sys.argv) == 1
    
    if is_full_schema_diagram:
```

Would rather us do something like this so we can remove the comment and is moar readable imo.

scripts/data_analysis/constants/common.py
luke-a-williams: No you're right, they should be all consistent. Will change.

scripts/data_analysis/constants/common.py
luke-a-williams: It's a tricky one I know what you mean, as very well accepted nomenclature for a dataframe is 'df', however it does break the rule of not being easily understood to all. I will change to data_frame to make it more explicit

scripts/data_analysis/generate_schema_diagram.py
luke-a-williams: Yes - was trying to think of the best way to do this, I like this!

scripts/data_analysis/README.md
luke-a-williams: I will change this and remove the mathematical nomenclature of nth argument too.

scripts/data_analysis/README.md
luke-a-williams: Agreed.

scripts/data_analysis/constants/common.py
luke-a-williams: I think both should be changed to child_schema actually, to make it more descriptive

chubberlisk (approved):
Wicked, thanks for addressing all my comments! I've been able to test and run the script successfully myself. 💯

PR #215, opened 2023-07-20
Update Probation Offender Search request body for searching

chubberlisk:
This should fix our smoke tests.


chubberlisk:
This is no longer needed as it's an issue with the documentation of Probation Offender Search. https://github.com/ministryofjustice/probation-offender-search/pull/599

PR #222, opened 2023-08-02
Offence information from nDelius

emileswarts:
This is a minimal integration with nDelius to pull back the description
of the main offence.

Pull requests to follow this:
  - The court date field will become a list
  - Check nullability of fields and remove where possible
  - Enable smoke tests for offences once API documentation is public
  - Combine main and additional offences
  - Include other fields to fill out the schema

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsintegrationapi/controllers/v1/person/OffencesController.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsintegrationapi/controllers/v1/person/OffencesController.kt
@@ -31,7 +31,11 @@ class OffencesController(
     val response = getOffencesForPersonService.execute(pncId)
 
     if (response.hasErrorCausedBy(UpstreamApiError.Type.ENTITY_NOT_FOUND, causedBy = UpstreamApi.NOMIS)) {
-      throw EntityNotFoundException("Could not find person with id: $pncId")
+      throw EntityNotFoundException("Could not find person with id: $pncId in NOMIS")
+    }
+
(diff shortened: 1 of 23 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsintegrationapi/controllers/v1/person/OffencesController.kt
```diff
@@ -30,8 +29,8 @@ class OffencesController(
     val pncId = encodedPncId.decodeUrlCharacters()
     val response = getOffencesForPersonService.execute(pncId)
 
-    if (response.hasErrorCausedBy(UpstreamApiError.Type.ENTITY_NOT_FOUND, causedBy = UpstreamApi.NOMIS)) {
-      throw EntityNotFoundException("Could not find person with id: $pncId")
+    if (response.hasError(UpstreamApiError.Type.ENTITY_NOT_FOUND)) {
+      throw EntityNotFoundException("Could not find offences for person with id: $pncId")
```
chubberlisk, line 33: I think the error message we had befoar was correct. If an upstream API returns 404, it means that the person with that ID couldn't be found i.e. they don't exist. The message above suggests to me that a person with the ID I gave was found but they had no offences. But we've said if a person is found and has no offences then we 200 and return an empty list.

chubberlisk (approved):
:shipit: 

Just one comment to be addressed!

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsintegrationapi/controllers/v1/person/OffencesController.kt
emileswarts: Cool yeah makes sense, updated.

PR #228, opened 2023-08-08
Update OpenAPI spec for offences

emileswarts:
OpenAPI spec update for Offences

### DIFF
--- a/openapi.yml
+++ b/openapi.yml
@@ -267,7 +267,7 @@ paths:
                   pagination:
                     $ref: "#/components/schemas/Pagination"
         "404":
-          description: Failed to find sentences a person with the provided PNC ID.
+          description: Failed to find offences a person with the provided PNC ID.
           content:
             application/json:
               schema:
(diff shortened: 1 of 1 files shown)


openapi.yml
```diff
@@ -603,12 +603,12 @@ components:
...
-        courtDate:
-          type: string
+          description: Criminal Justice System offence code, only present for Offences retrieved from Prison API
+        courtDates:
+          type: array
           format: date
-          example: 2018-02-10
-          description: Latest court date associated with the offence
+          example: [2018-02-10, 2019-02-10]
+          description: Court dates associated with offences. Offences from Prison API will always be a list of 1 item. Offences from nDelius will have more than 1 court date.
```
luke-a-williams, line 611: `type: array` requires you to specify an items sibling, so what you've put here won't render.
Consider something like:
```
        courtDates:
          type: array
          minItems: 1
          description: Court dates associated with offences. Offences from Prison API will always be a list of 1 item. Offences from nDelius will have more than 1 court date.
          items:
            type: string
            format: date
            example: [2018-02-10, 2019-02-10]
```

luke-a-williams (changes requested):
courtDates field needs amending

luke-a-williams (approved):
gtm

PR #234, opened 2023-08-11
Add active field to person sentences

popey2700:
#### Work done
- Added isActive to Hmpps-Integration-Api Sentence model from both Nomis and nDelius
- Created a mapping function for Nomis sentenceStatus String to Boolean

#### Tests
Does this PR include tests for the changes introduced?

- [x] Yes
- [ ] No, I found it difficult to test
- [ ] No, they are not required for this change


src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsintegrationapi/models/nomis/Sentence.kt
```diff
@@ -5,8 +5,17 @@ import uk.gov.justice.digital.hmpps.hmppsintegrationapi.models.Sentence as Integ
...
 data class Sentence(
   val sentenceDate: LocalDate?,
+  val sentenceStatus: String?,
 ) {
   fun toSentence(): IntegrationApiSentence = IntegrationApiSentence(
     dateOfSentencing = this.sentenceDate,
+    isActive = toIsActive(this.sentenceStatus),
   )
 }
+private fun toIsActive(sentenceStatus: String?): Boolean? {
```
luke-a-williams, line 15: This function name hurts my OCD. The question "isActive?", returning a boolean of true of false in certain cases of sentenceStatus, is already sufficient for this as a function name, no?
toIsActive implies we are changing the state of something, but we aren't doing this in my head?

src/test/kotlin/uk/gov/justice/digital/hmpps/hmppsintegrationapi/controllers/v1/SentencesControllerTest.kt
```diff
@@ -70,10 +72,13 @@ internal class SentencesControllerTest(
           """
           [
             {
-              "dateOfSentencing": "1990-01-01"
+              "dateOfSentencing": "1990-01-01",
+              "isActive": true
```
luke-a-williams, line 76: Should we also have a test for `"isActive": null`? Why exclude it?

src/test/kotlin/uk/gov/justice/digital/hmpps/hmppsintegrationapi/gateways/ndelius/GetSentencesForPersonTest.kt
```diff
@@ -65,9 +66,11 @@ class GetSentencesForPersonTest(
           listOf(
             generateTestSentence(
               dateOfSentencing = LocalDate.parse("2009-07-07"),
+              isActive = false,
             ),
             generateTestSentence(
               dateOfSentencing = LocalDate.parse("2009-09-01"),
+              isActive = false,
```
luke-a-williams, line 73: Is there something significant about the date being different whilst isActive remains the same? Otherwise should this be again testing all of the possible values (true, false, null)?

luke-a-williams (changes requested):

src/test/kotlin/uk/gov/justice/digital/hmpps/hmppsintegrationapi/models/nomis/PersonSentencesTest.kt
```diff
@@ -9,19 +9,47 @@ import uk.gov.justice.digital.hmpps.hmppsintegrationapi.models.nomis.Sentence as
...
+          sentenceStatus = "A",
         )
 
         val integrationApiSentence = nomisSentence.toSentence()
+
+        integrationApiSentence.dateOfSentencing.shouldBe(nomisSentence.sentenceDate)
+        integrationApiSentence.isActive.shouldBe(true)
+      }
+
+      it("toIsActive correctly maps 'I' to false") {
```
luke-a-williams, line 25: You changed the function name so it needs to be changed in the test name too

luke-a-williams (changes requested):

luke-a-williams (approved):
gtm

PR #237, opened 2023-08-21
Handle person not found in upstream search APIs

emileswarts:
When an upstream API returns 200 but no results for a person lookup, we need to stop further API calls with null values.

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsintegrationapi/services/GetOffencesForPersonService.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsintegrationapi/services/GetOffencesForPersonService.kt
@@ -33,7 +33,7 @@ class GetOffencesForPersonService(
 
     var nomisOffences: Response<List<Offence>> = Response(data = emptyList())
 
-    if(nomisNumber != null) {
+    if (nomisNumber != null) {
       nomisOffences = nomisGateway.getOffencesForPerson(nomisNumber)
 
       if (nomisOffences.errors.isNotEmpty()) {
@@ -43,7 +43,7 @@ class GetOffencesForPersonService(
 
     var nDeliusOffences: Response<List<Offence>> = Response(data = emptyList())
 
-    if(deliusCrn != null) {
+    if (deliusCrn != null) {
       nDeliusOffences = nDeliusGateway.getOffencesForPerson(deliusCrn)
 
       if (nDeliusOffences.errors.isNotEmpty()) {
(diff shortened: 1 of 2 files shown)


src/test/kotlin/uk/gov/justice/digital/hmpps/hmppsintegrationapi/services/GetOffencesForPersonServiceTest.kt
```diff
@@ -78,6 +78,32 @@ internal class GetOffencesForPersonServiceTest(
...
+          errors = emptyList(),
+        ),
+      )
+
+      val result = getOffencesForPersonService.execute(pncId)
+
+      result.shouldBe(Response(data = listOf(probationOffence1, probationOffence2, probationOffence3)))
+    }
+
+    it("Doesn't return probation offences when Probation Offender Search couldn't find the person by PNC ID") {
```
chubberlisk, line 94:
```suggestion
    it("returns only prison offences when the person cannot be found in Probation Offender Search") {
```

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsintegrationapi/services/GetOffencesForPersonService.kt
```diff
@@ -28,15 +28,27 @@ class GetOffencesForPersonService(
...
 
-    if (nomisOffences.errors.isNotEmpty()) {
-      return Response(emptyList(), nomisOffences.errors)
+    var nomisOffences: Response<List<Offence>> = Response(data = emptyList())
+
+    if (nomisNumber != null) {
+      nomisOffences = nomisGateway.getOffencesForPerson(nomisNumber)
+
+      if (nomisOffences.errors.isNotEmpty()) {
+        return Response(emptyList(), nomisOffences.errors)
```
chubberlisk, line 40: Not sure we should be returning here, what if offences can be found in Delius?

I think we agreed to surface all the errors from the service and let the controller decide what it wants to do with them?

chubberlisk (commented):
Agree with your suggestion on Slack about creating a new service to do the person lookup, something can return all the identifiers for a person sounds like a good future PR and improvement!

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsintegrationapi/services/GetOffencesForPersonService.kt
```diff
@@ -19,26 +19,26 @@ class GetOffencesForPersonService(
...
-    }
-
-    return Response(data = nomisOffences.data + nDeliusOffences.data)
+    return Response(
+      data = nomisOffences.data +
+        nDeliusOffences.data,
+      errors = responseFromPrisonerOffenderSearch.errors +
+        responseFromProbationOffenderSearch.errors +
+        nomisOffences.errors +
+        nDeliusOffences.errors,
```
chubberlisk, line 41: Just missing a unit test for dis now! 😉

chubberlisk (approved):
Pretty much there, just missing a unit test or two for combining the errors. Feel free to merge afterwards!

PR #239, opened 2023-08-22
OpenAPI spec for Alerts endpoint

emileswarts:


openapi.yml
```diff
@@ -276,6 +276,54 @@ paths:
...
+          required: false
+          description: The page number (starting from 1)
+        - in: query
+          name: per_page
+          schema:
+            type: number
+            minimum: 1
+            default: 10
+          required: false
+          description: The maximum number of results for a page
```
chubberlisk, line 302: Thinking these pagination query parameters have now reached my duplication threshold, could we refactor these like `EncodedPncId` in the line above?

chubberlisk (approved):
:shipit: 

Just one refactoring suggestion!

PR #240, opened 2023-08-22
Refactor: Use Person Service instead of calling individual gateways

emileswarts:
Re-use existing service to look up a person in upstream APIs. This approach is more cohesive and hides implementation details.

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsintegrationapi/services/GetOffencesForPersonService.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsintegrationapi/services/GetOffencesForPersonService.kt
@@ -4,23 +4,20 @@ import org.springframework.beans.factory.annotation.Autowired
 import org.springframework.stereotype.Service
 import uk.gov.justice.digital.hmpps.hmppsintegrationapi.gateways.NDeliusGateway
 import uk.gov.justice.digital.hmpps.hmppsintegrationapi.gateways.NomisGateway
-import uk.gov.justice.digital.hmpps.hmppsintegrationapi.gateways.PrisonerOffenderSearchGateway
-import uk.gov.justice.digital.hmpps.hmppsintegrationapi.gateways.ProbationOffenderSearchGateway
 import uk.gov.justice.digital.hmpps.hmppsintegrationapi.models.Offence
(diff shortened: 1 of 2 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsintegrationapi/services/GetOffencesForPersonService.kt
```diff
@@ -4,23 +4,20 @@ import org.springframework.beans.factory.annotation.Autowired
...
+  @Autowired val getPersonService: GetPersonService,
 ) {
   fun execute(pncId: String): Response<List<Offence>> {
-    val responseFromPrisonerOffenderSearch = prisonerOffenderSearchGateway.getPersons(pncId = pncId)
-    val responseFromProbationOffenderSearch = probationOffenderSearchGateway.getPerson(pncId = pncId)
-    val nomisNumber = responseFromPrisonerOffenderSearch.data.firstOrNull()?.identifiers?.nomisNumber
-    val deliusCrn = responseFromProbationOffenderSearch.data?.identifiers?.deliusCrn
+    val personResponse = getPersonService.execute(pncId = pncId)
+    val nomisNumber = personResponse.data["prisonerOffenderSearch"]?.identifiers?.nomisNumber
+    val deliusCrn = personResponse.data["probationOffenderSearch"]?.identifiers?.deliusCrn
```
chubberlisk, line 19: I think using a combination of [our UpstreamApi enum](https://github.com/ministryofjustice/hmpps-integration-api/blob/main/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsintegrationapi/models/UpstreamApi.kt) and overriding the `toString()` method or adding a new function, we could remove these magic strings in the `GetPersonService` and in this service.

However, I can live with this for now as I imagine we'll be removing these keys at some point for this service - I leave the decision in your capable hands!

chubberlisk (approved):
Definitely an improvement!

I slightly worry about using the existing `GetPersonService` (instead of creating a very slimmed down version). I can imagine it growing to include lots more attributes than we actually need for this use case i.e. just the `identifiers` we're interested in as well as the response object changing soon...

That being said, I'm taking the make it work, make it right, make it fast approach with this so it's not a big deal now!

🚢

PR #241, opened 2023-08-29
Add custodial flag to sentence model

emileswarts:
isCustodial will always be true for NOMIS sentences, nDelius has a non-nullable flag to indicate this.
Future PR will introduce a test helper to reduce some of the duplication in the unit tests.


chubberlisk (approved):
Love it - a nice and smol PR! 🚢

PR #242, opened 2023-08-30
Reduce duplication in tests for creating Sentences

emileswarts:
Introduce helper function with defaults that allows tests to only specify interesting fields to be asserted on.

Disable test that used to convert nDelius sentence length to terms, we will take a different approach in a following PR.


chubberlisk (approved):
Noice, noice, noice.

PR #245, opened 2023-08-31
Add PrisonTermCode to terms

emileswarts:
This code from Nomis indicates whether it's a custodial or license sentence


src/test/kotlin/uk/gov/justice/digital/hmpps/hmppsintegrationapi/models/nomis/PersonSentencesTest.kt
```diff
@@ -122,65 +128,7 @@ class PersonSentencesTest : DescribeSpec(
...
-        val integrationApiSentence = nomisSentence.toSentence()
-
-        integrationApiSentence.terms.shouldBe(
-          listOf(
-            IntegrationApiTerm(
-              years = 3,
-              months = 9,
-              weeks = null,
-              days = null,
-              hours = null,
```
emileswarts, line 183: Found duplicated tests, removed, guessing a merge gone awry

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsintegrationapi/models/Term.kt
```diff
@@ -6,4 +6,5 @@ data class Term(
   val weeks: Int? = null,
   val days: Int? = null,
   val hours: Int? = null,
+  val prisonTermCode: String? = null,
```
chubberlisk, line 9: Part of me feels like just `code` is fine since we're within a `Term`, not sure...

emileswarts: Yeah I can see that..

emileswarts: Maybe the "prison" part adds valuable context? 
Assume careful consideration went into this naming but let me take a note and we can discuss afterwards as a refactor

PR #247, opened 2023-08-31
Update sentence length restructure

emileswarts:
Since we cannot cleanly combine community and custodial sentences, we're changing the structure to accommodate both, while preserving the context. Introduce a length object which has "duration" for community sentences, and "terms" for custodial sentences.

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsintegrationapi/models/Length.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsintegrationapi/models/Length.kt
@@ -1,7 +1,0 @@
-package uk.gov.justice.digital.hmpps.hmppsintegrationapi.models
-
-data class Length(
-  val duration: Int? = null,
-  val units: String? = null,
-  val terms: List<Term> = listOf(Term()),
-)
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsintegrationapi/models/Length.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsintegrationapi/models/Length.kt
@@ -1,0 +1,7 @@
+package uk.gov.justice.digital.hmpps.hmppsintegrationapi.models
+
+data class Length(
+  val duration: Int? = null,
+  val units: String? = null,
+  val terms: List<Term> = listOf(Term()),
+)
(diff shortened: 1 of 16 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsintegrationapi/models/Length.kt
```diff
@@ -0,0 +1,7 @@
+package uk.gov.justice.digital.hmpps.hmppsintegrationapi.models
+
+data class Length(
+  val duration: Int? = null,
+  val units: String? = null,
+  val terms: List<Term> = listOf(Term()),
```
chubberlisk:
```suggestion
  val terms: List<Term> = emptyList(),
```
If there are no terms, then I don't think we should be defaulting to a list with a single object with all null properties...

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsintegrationapi/models/SentenceLength.kt
```diff
@@ -1,8 +1,7 @@
 package uk.gov.justice.digital.hmpps.hmppsintegrationapi.models
 
 data class SentenceLength(
-  val days: Int? = null,
-  val weeks: Int? = null,
-  val months: Int? = null,
-  val years: Int? = null,
+  val duration: Int? = null,
+  val units: String? = null,
+  val terms: List<SentenceTerm> = listOf(SentenceTerm()),
```
chubberlisk, line 6:
```suggestion
  val terms: List<SentenceTerm> = emptyList(),
```

If there are no terms, then I don't think it makes sense to return a single object with null for all properties...

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsintegrationapi/models/ndelius/Sentence.kt
```diff
@@ -1,21 +1,18 @@
...
-      "months" -> listOf(IntegrationApiTerm(months = this.length))
-      "weeks" -> listOf(IntegrationApiTerm(weeks = this.length))
-      "days" -> listOf(IntegrationApiTerm(days = this.length))
-      "hours" -> listOf(IntegrationApiTerm(hours = this.length))
-      else -> listOf(IntegrationApiTerm())
-    }
+  fun toLength(): IntegrationApiLength {
+    return IntegrationApiLength(
+      duration = this.length,
+      units = this.lengthUnits?.lowercase(),
```
chubberlisk, line 14: I'm wondering whether we should keep it as uppercase. In our images endpoint for enum values like view and orientation they are uppercase. Wondering if our convention should be if it's an enum value, it should be uppercase?

chubberlisk (approved):
🚢

PR #250, opened 2023-09-01
Openapi spec sentences

emileswarts:
Add tags to indicate upstream APIs for additional context.

### DIFF
--- a/openapi.yml
+++ b/openapi.yml
@@ -824,8 +824,8 @@ components:
           type: array
           minItems: 0
           items:
-            $ref: "#/components/schemas/Term"
-    Term:
+            $ref: "#/components/schemas/SentenceTerm"
+    SentenceTerm:
       type: object
       properties:
         years:
--- a/openapi.yml
+++ b/openapi.yml
@@ -255,6 +255,40 @@ paths:
                 PersonNotFoundError:
                   $ref: "#/components/examples/PersonNotFoundError"
 
+  /v1/persons/{Id}/sentences:
+    get:
+      tags:
+        - persons
+      summary: Returns sentences associated with a person.
+      parameters:
+        - $ref: "#/components/parameters/EncodedPncId"
+        - $ref: "#/components/parameters/Page"
+        - $ref: "#/components/parameters/PerPage"
+      responses:
+        "200":
+          description: Successfully found sentences for a person with the provided PNC ID.
+          content:
+            application/json:
+              schema:
+                type: object
+                properties:
+                  data:
+                    type: array
+                    minItems: 0
+                    items:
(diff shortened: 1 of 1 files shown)


openapi.yml
```diff
@@ -255,6 +255,40 @@ paths:
...
+                properties:
+                  data:
+                    type: array
+                    minItems: 0
+                    items:
+                      $ref: "#/components/schemas/Sentence"
+                  pagination:
+                    $ref: "#/components/schemas/Pagination"
+        "404":
+          description: Failed to find sentences a person with the provided PNC ID.
```
chubberlisk, line 283:
```suggestion
          description: Failed to find sentences for a person with the provided PNC ID.
```

openapi.yml
```diff
@@ -743,6 +769,106 @@ components:
...
           example: 2008/0545166T
           description: An identifier from the Police National Computer (PNC)
+    Sentence:
+      type: object
+      properties:
+        dataSource:
+          type: string
+          enum: [NOMIS, NDELIUS]
+          example: NOMIS
+          description: Which upstream API the sentence originates from
```
chubberlisk, line 779:
```suggestion
          description: Which legacy system the sentence originates from
```

Thinking "upstream API" would be "Prison API" and "legacy system" is "NOMIS"?

openapi.yml
```diff
@@ -743,6 +769,106 @@ components:
...
+          tags: nomis, ndelius
+          type: string
+          example: Young Offender Inst - >=12 mths
+          description: Description of the sentence
+        isActive:
+          tags: nomis, ndelius
+          type: boolean
+          example: true
+          description: Whether the sentence is active
+        isCustodial:
```
chubberlisk, line 796: Is this missing tags?

openapi.yml
```diff
@@ -743,6 +769,106 @@ components:
...
+          enum: [Hours, Days, Weeks, Months, Years]
+          example: Hours
+          description: Time unit that is used in combination with the duration field
+        terms:
+          tags: nomis
+          type: array
+          minItems: 0
+          items:
+            $ref: "#/components/schemas/Term"
+    Term:
```
chubberlisk, line 828: Should this be `SentenceTerm`?

chubberlisk (approved):
🚢

openapi.yml
emileswarts: Intentionally left this one out. 
The flag exists in nDelius only, we always set this to true for Nomis within our own API.

Feels like tagging both, or only one API in this case would be deceiving.
Thoughts?

chubberlisk: Ahhhhh okay, I think let's leave it out!

PR #256, opened 2023-09-13
Replace PNC Id with HMPPS Id in OpenAPI spec

emileswarts:
This is a placeholder and will become the new global identifier once it's ready.

### DIFF
--- a/openapi.yml
+++ b/openapi.yml
@@ -60,16 +60,16 @@ paths:
                 NoQueryParametersBadRequestError:
                   $ref: "#/components/examples/NoQueryParametersBadRequestError"
 
-  /v1/persons/{encodedPncId}:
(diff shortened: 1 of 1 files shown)


openapi.yml
```diff
@@ -394,9 +394,9 @@ paths:
 
 components:
   parameters:
-    EncodedPncId:
-      name: encodedPncId
-      description: A URL-encoded Police National Computer (PNC) identifier.
+    HmppsId:
+      name: hmppsId
+      description: A URL-encoded Police National Computer (HMPPS) identifier.
```
AprilMD, line 399: This description still says Police National Computer - should probably just say "A URL-encoded HMPPS identifier."

AprilMD (approved):

PR #259, opened 2023-09-15
Hia 505 person risk pull back one field

AprilMD:
## Context

Several risk assessment endpoints are coming up. We want to make implementing these endpoints easier and as a result we begin by taking a single field from an upstream api. All of these endpoints will use ARN and as a result this PR includes the ARN gateway.

Worth noting that this PR does not include smoke tests as the api docs for ARN are not public. We are requesting/exploring making them public in this [ticket](https://dsdmoj.atlassian.net/jira/software/c/projects/HIA/boards/1056?modal=detail&selectedIssue=HIA-504)

## Changes proposed in this PR

- Creation of a ARN gateway
- Implement RiskPredictorScore Models, Service and controller

PR #260, opened 2023-09-20
Implement rest of risk assessment score model

popey2700:
## Context

The original risk assessment gateway, service and controller was created in [PR 505](https://github.com/ministryofjustice/hmpps-integration-api/pull/259). This PR builds upon the work done adding the remaining golden record risk fields.

Worth noting that this PR does not include smoke tests as the api docs for ARN are not public. We are requesting/exploring making them public in this [ticket](https://dsdmoj.atlassian.net/jira/software/c/projects/HIA/boards/1056?modal=detail&selectedIssue=HIA-504)

- Implementation of risk scoring golden record
- Updating openApi spec with new endpoint
- Changed General predictor score to be a String rather than an Int


chubberlisk (approved):
Noiceeeeeee, just a couple of comments around the OpenAPI spec. Other than that, all good!

PR #261, opened 2023-09-21
Create Get Person Risks basic end-point

AprilMD:
Create Get Persons Risks (/v1/persons/{hmppsId}/risks) endpoint which exposes Risk of Serious Harm (RoSH) data from the Assess Risks and Needs API.

Smoke tests to follow.


src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsintegrationapi/gateways/AssessRisksAndNeedsGateway.kt
```diff
@@ -10,7 +10,9 @@ import uk.gov.justice.digital.hmpps.hmppsintegrationapi.models.Response
 import uk.gov.justice.digital.hmpps.hmppsintegrationapi.models.UpstreamApi
 import uk.gov.justice.digital.hmpps.hmppsintegrationapi.models.UpstreamApiError
 import uk.gov.justice.digital.hmpps.hmppsintegrationapi.models.RiskPredictorScore as IntegrationAPIRiskPredictorScore
+import uk.gov.justice.digital.hmpps.hmppsintegrationapi.models.Risks as IntegrationApiRisk
```
popey2700, line 13: I believe (Not certain) we've used IntegrationAPI opposed to IntegrationApi in the codebase as our alias convention - we should agree upon one and stick to it. I prefer IntegrationApi how you have it.

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsintegrationapi/models/assessRisksAndNeeds/Risks.kt
```diff
@@ -0,0 +1,12 @@
+package uk.gov.justice.digital.hmpps.hmppsintegrationapi.models.assessRisksAndNeeds
+
+import java.time.LocalDateTime
+import uk.gov.justice.digital.hmpps.hmppsintegrationapi.models.Risks as IntegrationApiRisks
+
+data class Risks(
+  val assessedOn: LocalDateTime? = null,
```
popey2700, line 7: Can this be of type LocalDateTime straight away or does it need to be a string and then parsed?

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsintegrationapi/gateways/AssessRisksAndNeedsGateway.kt
AprilMD: We've used a mix in the codebase. You'll also be glad to hear, we've actually [got a ticket in the backlog](https://dsdmoj.atlassian.net/browse/HIA-426) now to create a convention for these with a few different options (all of which, I believe, are better than either IntegrationAPI/Api).

src/test/kotlin/uk/gov/justice/digital/hmpps/hmppsintegrationapi/models/assessRisksAndNeeds/RisksTest.kt
```diff
@@ -0,0 +1,24 @@
...
+
+import io.kotest.core.spec.style.DescribeSpec
+import io.kotest.matchers.shouldBe
+import java.time.LocalDateTime
+import uk.gov.justice.digital.hmpps.hmppsintegrationapi.models.assessRisksAndNeeds.Risks as ArnRisks
+
+class RisksTest : DescribeSpec(
+  {
+    describe("#toRisks") {
+      it("maps ARN Risks to Integration API Risks") {
```
popey2700, line 11: This is pretty picky so not fussed normal test naming convention for 1-1 attributes is: "maps one-to-one attributes to x attributes"

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsintegrationapi/gateways/AssessRisksAndNeedsGateway.kt
popey2700: Sensational, thanks for the heads up!

popey2700 (approved):
LGTM!

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsintegrationapi/models/assessRisksAndNeeds/Risks.kt
AprilMD: Yes, it can be of type LocalDateTime straight away.

The webclient runs bodyToMono on the response from the upstream API, which converts it from json into an object (in this case, a Risks object). 

https://github.com/ministryofjustice/hmpps-integration-api/blob/3090c821fc6f0b63f2b1a11da5d1a921cc46fec5/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsintegrationapi/extensions/WebClientWrapper.kt#L33-L36

Our gateway tests implicitly test for this, as the mock server stubs the date-time as JSON string.

https://github.com/ministryofjustice/hmpps-integration-api/blob/4b0eec6a479576168c09bdd7fa9e31ec4130678f/src/test/kotlin/uk/gov/justice/digital/hmpps/hmppsintegrationapi/gateways/assessrisksandneeds/GetRisksForPersonTest.kt#L37-L47

src/test/kotlin/uk/gov/justice/digital/hmpps/hmppsintegrationapi/models/assessRisksAndNeeds/RisksTest.kt
AprilMD: Ty done

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsintegrationapi/models/assessRisksAndNeeds/Risks.kt
popey2700: Good points this makes sense, I wonder if there are unnecessary parses in the codebase where we read time related fields in as Strings when we don't need to. Will make a note of this!

PR #266, opened 2023-09-27
Created needs model with a single field for thin slice

popey2700:
## Context

TODO: Future Alex type context here

Worth noting that this PR does not include smoke tests as the api docs for ARN are not public. We are requesting/exploring making them public in this [ticket](https://dsdmoj.atlassian.net/jira/software/c/projects/HIA/boards/1056?modal=detail&selectedIssue=HIA-504)

- Implement Needs Models, Service and controller

### DIFF
--- a/src/test/kotlin/uk/gov/justice/digital/hmpps/hmppsintegrationapi/models/assessRisksAndNeeds/NeedsTest.kt
+++ b/src/test/kotlin/uk/gov/justice/digital/hmpps/hmppsintegrationapi/models/assessRisksAndNeeds/NeedsTest.kt
@@ -1,0 +1,22 @@
+package uk.gov.justice.digital.hmpps.hmppsintegrationapi.models.assessRisksAndNeeds
+
+import io.kotest.core.spec.style.DescribeSpec
+import io.kotest.matchers.shouldBe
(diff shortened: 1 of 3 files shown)


src/test/kotlin/uk/gov/justice/digital/hmpps/hmppsintegrationapi/models/assessRisksAndNeeds/NeedsTest.kt
```diff
@@ -0,0 +1,22 @@
...
+
+import io.kotest.core.spec.style.DescribeSpec
+import io.kotest.matchers.shouldBe
+import java.time.LocalDateTime
+import uk.gov.justice.digital.hmpps.hmppsintegrationapi.models.assessRisksAndNeeds.Needs as ArnNeeds
+
+class NeedsTest : DescribeSpec(
+  {
+    describe("#toNeeds") {
+      it("maps ARN Needs to integration API Needs") {
```
AprilMD, line 11: Usually for 1:1s we use "maps one-to-one attributes to Integration API attributes"

AprilMD (approved):
fab

PR #272, opened 2023-10-04
updated the diagram to include a meaningful ARN section

luke-a-williams:
Updated the legacy containers section with the new API, but also reworded the text to be "Some" of the APIS will be replaced, not to cause confusion with ones that are persisting and just evolving.


popey2700 (approved):
Lovely stuff

PR #270, opened 2023-10-03
Add Risk Summary to Risks

AprilMD:
## Context

In #261, we created the thin slice for the `/v1/persons/{id}/risks` endpoint which will return the risk to serious harm (RoSH) risks.

Our golden record includes four top-level properties:

1. `riskToSelf` - added in #262 
2. `otherRisks` - added in #263
3. `summary` - **added in this PR**
4. `assessedOn` - added in #261 (thin slice)

## Changes proposed in this pull request

- Add models for Assess Risks and Needs (ARN) and our API to return `riskSummary`. 
    - Some of these attributes map one to one to our golden record so no transformation is done.
    - The Risk in Community and Risk in Context attributes have been have been mapped onto a restructured model with person groups as keys and risk levels as values. This mapping is done via a Risk in Context function.

## Next steps

1. Documentation

PR #278, opened 2023-10-17
Include rosh risks smoke test

popey2700:
## Context

We completed ROSH risks in this [PR](https://github.com/ministryofjustice/hmpps-integration-api/pull/270). However at the time we did not introduce smoke tests due to various reasons such as bugs with our smoke tests and ARN documentation not being public. These problems are now resolved and as a result the smoke test can be properly introduced.

- Updated and included ROSH smoke tests.


chubberlisk (approved):
:shipit:

PR #281, opened 2023-10-19
Hia 566 implement hmpps id property to person response

popey2700:
## Context

With the project pivoting to use a hmppsId its important we update our initial entry endpoint persons response to contain a hmpps property. The actual hmppsId value will initially be a pnc number until our migration to using crn’s is complete. Then the hmppsId value will be a crn. Further, once the ‘official’ hmppsId is implemented the hmppsId will just be the actual hmpps value.

- Persons endpoint now returns a hmppsId in its response


chubberlisk (approved):
:shipit:

PR #282, opened 2023-10-23
Began updating ImageMetaData endpoint to use probabtion offender gateway

popey2700:


src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsintegrationapi/controllers/v1/person/PersonController.kt
```diff
@@ -63,18 +63,18 @@ class PersonController(
     return response.data
   }
 
-  @GetMapping("{encodedPncId}/images")
+  @GetMapping("{encodedHmppsId}/images")
```
emileswarts, line 66: Wonder if we need the "encoded" part given the new value won't be encoded

src/test/kotlin/uk/gov/justice/digital/hmpps/hmppsintegrationapi/gateways/probationoffendersearch/ProbationOffenderSearchGatewayTest.kt
```diff
@@ -221,45 +328,46 @@ class ProbationOffenderSearchGatewayTest(
...
+          "[]",
+        )
 
-      val response = probationOffenderSearchGateway.getPerson(pncId)
+        val response = probationOffenderSearchGateway.getPerson(hmppsId)
 
-      response.data.shouldBeNull()
+        response.data.shouldBeNull()
+      }
     }
```
emileswarts, line 371: Tempted to put these new tests in different files so it's easy to clean up when we've made the switch

chubberlisk: We decided that having a `describe` block for the feature flag was enough.

chubberlisk (approved):

PR #287, opened 2023-11-01
Update NOMIS Gateway to get latest sentence key dates and adjustments

chubberlisk:
## Context

We want to build an endpoint to return the latest sentence key dates and adjustments i.e. prisoner release dates and adjustments.

First, we're going to build a thin slice of the API endpoint and then flesh out the response. As we're going to need to call two API endpoints from NOMIS (via Prison API), this PR has been split at this point and will have a follow-up PR that add the service and controller bits later.

## Changes proposed in this pull request

- Add two new methods within our NOMIS gateway to return the latest sentence key dates and adjustments.
- Add models for only returning one field for sentence key dates and adjustments.

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsintegrationapi/models/nomis/SentenceKeyDates.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsintegrationapi/models/nomis/SentenceKeyDates.kt
@@ -1,0 +1,13 @@
+package uk.gov.justice.digital.hmpps.hmppsintegrationapi.models.nomis
+
+import uk.gov.justice.digital.hmpps.hmppsintegrationapi.models.SentenceKeyDate
+import uk.gov.justice.digital.hmpps.hmppsintegrationapi.models.SentenceKeyDates
+import java.time.LocalDate
+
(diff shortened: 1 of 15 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsintegrationapi/models/nomis/SentenceKeyDates.kt
```diff
@@ -0,0 +1,13 @@
+package uk.gov.justice.digital.hmpps.hmppsintegrationapi.models.nomis
+
+import uk.gov.justice.digital.hmpps.hmppsintegrationapi.models.SentenceKeyDate
+import uk.gov.justice.digital.hmpps.hmppsintegrationapi.models.SentenceKeyDates
```
emileswarts, line 4: Looks like a double import here

chubberlisk: Ahhhhhh, they're different classes. The idea is that `SentenceKeyDate` can be reused for the different key dates properties as they will all have a date, calculated date, override date etc.

image

image

PR #332, opened 2023-12-15
HIA-176/HIA-388 -- Refactoring ProbationOffenderSearchGateway class code.

chiaramapellimt:


src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsintegrationapi/gateways/ProbationOffenderSearchGateway.kt
```diff
@@ -36,19 +36,22 @@ class ProbationOffenderSearchGateway(@Value("\${services.probation-offender-sear
 
     return when (result) {
       is WebClientWrapperResponse.Success -> {
-        if (result.data.isEmpty()) {
-          Response(
-            data = null,
-            errors = listOf(
+        val persons = result.data
+        val person = persons.firstOrNull()?.toPerson()
```
emileswarts, line 40: Is persons.first()?.toPerson()
the same as:
persons.firstOrNull()?.toPerson()

emileswarts (approved):

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsintegrationapi/gateways/ProbationOffenderSearchGateway.kt
chiaramapellimt: No it's a bit different, the first expression returns the the first element of the persons list, and if the list is empty, it will throw a NoSuchElementException (so basically it doesn't cover the case when the list is empty). The second expression returns the first element of the persons list and I think it's safer to use in cases where the list might be empty.
I think using persons.firstOrNull()?.toPerson() is more defensive because it avoids null pointer exceptions, because if we get an empty list for persons, firstOrNull() will return null, and the rest of ?.toPerson() will safely handle the null case without throwing exceptions etc

PR #339, opened 2024-01-03
Enhanced the generate certs script so that if successful, it upload to S3

chiaramapellimt:
Enhanced the generate certs script so that if successful, it uploads to S3 bucket.


popey2700 (approved):
LGTM! Nice work:)

PR #341, opened 2024-01-09
Fix broken urls data scripts

luke-a-williams:
I'm not fully sure why, but an external member of MOJ to our team changed the urls here, but not all of them affected by the change

I've also put some logic in to more obviously communicate when a json response is empty.


CslBing (approved):

luke-a-williams:
Smoke tests were unaffected by these changes, so merged

PR #342, opened 2024-01-09
ADR: Certificates in production

popey2700:
## Context

This PR is for creating an ADR to explain the design decision to use certificates for API security in production.

- Added ADR 0008 ADR


CslBing (approved):
looks good

PR #366, opened 2024-02-06
HIA-390-2

AlexHerbertMoJ:
Removing and ignoring an erroneously committed file
Updating OpenAPI spec to reflect changes to the API which was missed


popey2700 (approved):
LGTM!

PR #369, opened 2024-02-07
HIA 661 persons search 500 with pnc bug

popey2700:
## Context

When searching persons with PNC if the probation offender search doesn’t return a person the request will crash into a 500 rather than a 404. As a result of Prisoner search not taking a PNC we decided to change functionality so if someone searches with PNC they will only get ProbationOffender results back

- Changed PNC search flow


src/test/kotlin/uk/gov/justice/digital/hmpps/hmppsintegrationapi/gateways/prisoneroffendersearch/PrisonerOffenderSearchGatewayTest.kt
```diff
@@ -108,42 +106,13 @@ class PrisonerOffenderSearchGatewayTest(
...
 
-      val response = prisonerOffenderSearchGateway.getPersons("Obi-Wan", null, null, null)
+      val response = prisonerOffenderSearchGateway.getPersons("Obi-Wan", null, null)
 
       response.data.count().shouldBe(1)
       response.data.first().firstName.shouldBe("Obi-Wan")
       response.data.first().lastName.shouldBe("Kenobi")
     }
 
-    it("returns person(s) when searching on hmppsId only") {
```
AlexHerbertMoJ, line 118: Was this removed on purpose?

AlexHerbertMoJ (commented):
Not sure I'm the best person to review; I'd appreciate it if you had time to talk me through the changes you made and why they work

Additionally, I don't see a test case for the situation presented in the bug. Again, I'm still acclimatising to this team's ways of working, but would it not be good to make a test for the specific case to ensure it doesn't crop up again in the future?

src/test/kotlin/uk/gov/justice/digital/hmpps/hmppsintegrationapi/gateways/prisoneroffendersearch/PrisonerOffenderSearchGatewayTest.kt
CslBing: Yes, we have excluded pnc search via prisoner, as it is not a search term used by prisoner

AlexHerbertMoJ (approved):
Looks good to me :)

PR #375, opened 2024-02-27
HIA-678 - Expand persons needs endpoint

chiaramapellimt:
Expanding the persons' needs endpoint


AlexHerbertMoJ (commented):
The ticket shows that there should be a 'name' field here in addition to risk of harm, risk of reoffending, section/type and severity. Just wanted to make sure that these are not needed if they're not included.

AlexHerbertMoJ (approved):

PR #384, opened 2024-03-06
Golden record for assessments

chiaramapellimt:


src/test/kotlin/uk/gov/justice/digital/hmpps/hmppsintegrationapi/controllers/v1/person/RiskCategoriesControllerTest.kt
```diff
@@ -78,11 +78,21 @@ internal class RiskCategoriesControllerTest(
...
             "assessments": [
-            {
-                "classificationCode": "C"
-            }
-          ]
-        }
+                {
+                  "classificationCode": "C",
+                  "classification": null,
+                  "assessmentCode": null,
```
CslBing, line 84: Maybe we can have better test data than nulls?

PR #388, opened 2024-03-19
Change MappaDetail domain to use Number instead of Integer and update…

CslBing:
… openapi documentation

### DIFF
--- a/openapi.yml
+++ b/openapi.yml
@@ -432,18 +432,27 @@ paths:
                     properties:
                       level:
                         type: number
+                        example: 1
                       levelDescription:
                         type: string
+                        example: Description of M1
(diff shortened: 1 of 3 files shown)


openapi.yml
```diff
@@ -431,11 +431,11 @@ paths:
                     type: object
                     properties:
                       level:
-                        type: string
+                        type: integer
```
AlexHerbertMoJ, line 434: Any reason they're Integer here and Number below?

CslBing: No particular reason, as it would be the same either way. Updated to use number instead

AlexHerbertMoJ (approved):
LGTM

PR #390, opened 2024-03-28
Fix dependencies vulnerabilities issues

CslBing:


gradlew
```diff
@@ -80,13 +80,11 @@ do
     esac
 done
 
-APP_HOME=$( cd "${APP_HOME:-./}" && pwd -P ) || exit
-
-APP_NAME="Gradle"
+# This is normally unused
```
AlexHerbertMoJ, line 83: If it's normally unused and is commented, do we still need it?

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsintegrationapi/models/assessRisksAndNeeds/ArnRiskSummary.kt
```diff
@@ -12,19 +12,20 @@ data class ArnRiskSummary(
...
+      riskIncreaseFactors = this.riskIncreaseFactors,
+      riskMitigationFactors = this.riskMitigationFactors,
+      overallRiskLevel = this.overallRiskLevel,
+      riskInCommunity = if (!this.riskInCommunity.isNullOrEmpty()) toRiskInContext(this.riskInCommunity) else null,
+      riskInCustody = if (!this.riskInCustody.isNullOrEmpty()) toRiskInContext(this.riskInCustody) else null,
+    )
 
   private fun toRiskInContext(arnRiskInContext: Map<String, List<String>>): Map<String, String> {
-    var riskInContext = mutableMapOf<String, String>()
+    val riskInContext = mutableMapOf<String, String>()
```
AlexHerbertMoJ, line 28: If it's a mutable map, should this not be var and not val?

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsintegrationapi/models/hmpps/HmppsAuditEvent.kt
```diff
@@ -4,6 +4,7 @@ import java.time.ZoneId
 import java.time.ZonedDateTime
 import java.time.format.DateTimeFormatter
 
+@Suppress("PropertyName")
```
AlexHerbertMoJ, line 7: What's this for?

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsintegrationapi/models/hmpps/Response.kt
```diff
@@ -2,12 +2,13 @@ package uk.gov.justice.digital.hmpps.hmppsintegrationapi.models.hmpps
 
 data class Response<T>(val data: T, val errors: List<UpstreamApiError> = emptyList()) {
   companion object {
-    fun <T> merge(responses: List<Response<List<T>>>): Response<List<T>> =
-      Response(data = responses.flatMap { it.data }, errors = responses.flatMap { it.errors })
+    fun <T> merge(responses: List<Response<List<T>>>): Response<List<T>> = Response(data = responses.flatMap { it.data }, errors = responses.flatMap { it.errors })
```
AlexHerbertMoJ, line 5: Strange that all other files so far have expanded the code but this one has compressed it.

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsintegrationapi/models/assessRisksAndNeeds/ArnRiskSummary.kt
CslBing: val prevent reassign value to the variable. It does not prevent mutableMap to change it's content.

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsintegrationapi/models/hmpps/HmppsAuditEvent.kt
CslBing: kolint and IntelliJ has validation on property name to start with lower case. Audit event class require a property called "when", however, this is a keyword in kotlin, therefore we need to wrap the name "when" with back ticks, and it will break the name validation, therefore I have to suppress that for this class

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsintegrationapi/models/hmpps/Response.kt
CslBing: Kolint prefer assignment in one line when number of parameter is small.

src/main/resources/application-local-docker.yml
```diff
@@ -47,6 +47,7 @@ authorisation:
 hmpps.sqs:
   provider: localstack
   localstackUrl: http://local-stack-aws:4566
+  useWebToken: false
```
AlexHerbertMoJ, line 50: what's this for?

AlexHerbertMoJ (commented):
Some small questions but otherwise this looks ok

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsintegrationapi/models/hmpps/HmppsAuditEvent.kt
AlexHerbertMoJ: Cool

src/main/resources/application-local-docker.yml
CslBing: Version 3 of hmpps.sqs changed useWebToken to default to true, therefore we need to set it to false when we using localstack. [Breaking changes](https://github.com/ministryofjustice/hmpps-spring-boot-sqs/blob/main/release-notes/3.x.md#Breaking-changes)

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsintegrationapi/models/assessRisksAndNeeds/ArnRiskSummary.kt
AlexHerbertMoJ: Great answer, thanks

src/main/resources/application-local-docker.yml
AlexHerbertMoJ: Great

AlexHerbertMoJ (approved):
LGTM

PR #404, opened 2024-04-24
Hia 717 epf endpoint update

popey2700:
## Context

- Our upstreamApi source for our EPF endpoint was updated recently and as a result we needed to update our codebase and response

- Removed no longer available fields from EPF response
- Added new relevant fields to  EPF response


src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsintegrationapi/models/hmpps/CaseDetail.kt
```diff
@@ -19,16 +20,19 @@ data class Name(
 )
 
 data class CaseSentence(
-  val date: String? = null,
-  val sentencingCourt: SentencingCourt? = null,
```
CslBing, line 23: maybe change to mark as obsolete instead delete?

CslBing (approved):
LGTM

PR #410, opened 2024-05-08
Change open-api

chiaramapellimt:


openapi.yml
```diff
@@ -1902,7 +1865,85 @@ components:
           nullable: true
         prison:
           $ref: "#/components/schemas/Prison"
-
+    PrisonerOffenderSearch:
+      type: object
```
CslBing, line 1869: Why PrisonerOffenderSearch and ProbationOffenderSearch are two objects if they are the same?

chiaramapellimt: you're right, I needed to rekick the pipeline, I will squish them together now

CslBing (approved):
LGTM

PR #439, opened 2024-06-07
Standardising open api

chiaramapellimt:


openapi.yml
```diff
@@ -127,7 +127,7 @@ paths:
         - persons
       summary: Returns metadata of images associated with a person ordered by newest to oldest date.
       parameters:
-        - $ref: "#/components/parameters/HmppsId"
+        - $ref: "#/components/parameters/hmppsId"
         - $ref: "#/components/parameters/Page"
```
popey2700, line 131: Nice formalisation, my question is are we going to commit to all parameters being lowercase i.e Page and PerPage?

chiaramapellimt: we probs should too

popey2700 (approved):

PR #443, opened 2024-06-13
Improvements

chiaramapellimt:


openapi.yml
```diff
@@ -12,7 +12,7 @@ paths:
     get:
       tags:
         - persons
-      summary: Returns person(s) by search criteria, sorted by date of birth. At least one query parameter must be specified.
+      summary: Returns person(s) by search criteria, sorted by date of birth (descending). At least one query parameter must be specified.
```
CslBing, line 15: It might be more clearer to say newest first or oldest first instead of descending or ascending

PR #445, opened 2024-06-14
Refactor risk endpoints to return empty data when upstream returns 403

chiaramapellimt:
We decided to make the changes inside the gateway and the lower level functions to avoid overloading the service.

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsintegrationapi/services/GetRiskPredictorScoresForPersonService.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsintegrationapi/services/GetRiskPredictorScoresForPersonService.kt
@@ -5,6 +5,7 @@ import org.springframework.stereotype.Service
 import uk.gov.justice.digital.hmpps.hmppsintegrationapi.gateways.AssessRisksAndNeedsGateway
 import uk.gov.justice.digital.hmpps.hmppsintegrationapi.models.hmpps.Response
 import uk.gov.justice.digital.hmpps.hmppsintegrationapi.models.hmpps.RiskPredictorScore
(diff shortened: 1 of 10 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsintegrationapi/services/GetRiskSeriousHarmForPersonService.kt
```diff
@@ -19,6 +20,9 @@ class GetRiskSeriousHarmForPersonService(
 
     if (deliusCrn != null) {
       personRisks = assessRisksAndNeedsGateway.getRiskSeriousHarmForPerson(id = deliusCrn)
+      if (personRisks.hasError(UpstreamApiError.Type.FORBIDDEN)) {
```
CslBing, line 23: If we refactor AssessRisksAndNeedsGateway to return emptyList() instead of null, then we will not need to have this if statement any more

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsintegrationapi/services/GetRiskPredictorScoresForPersonService.kt
```diff
@@ -19,6 +20,9 @@ class GetRiskPredictorScoresForPersonService(
 
     if (deliusCrn != null) {
       personRiskPredictorScores = assessRisksAndNeedsGateway.getRiskPredictorScoresForPerson(id = deliusCrn)
+      if (personRiskPredictorScores.hasError(UpstreamApiError.Type.FORBIDDEN)) {
```
CslBing, line 23: This if statement is not required as it will return the same result as per line 24 when personRiskPredictorScores has error

CslBing (approved):

PR #448, opened 2024-06-26
Add new PES prisoner endpoint

garethmdavies:
New endpoint to support the HMPPS Prisoner Education programme, who need to automatically provision new Azure accounts for prisoners as they move into non-Launchpad sites.


CslBing:
Sorry for late review, but the changes looked good.

PR #452, opened 2024-07-03
PES - add hmpps id endpoint for a given nomis number

garethmdavies:
Add endpoint to retrieve HMPPS ID given a nomis number. Until the HMPPS core person work is finialised this endpoint will return the crn in most cases but could also return the nomis number where no crn exists for a person.

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsintegrationapi/models/probationoffendersearch/Offender.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsintegrationapi/models/probationoffendersearch/Offender.kt
@@ -33,7 +33,7 @@ data class Offender(
           deliusCrn = otherIds.crn,
         ),
       pncId = otherIds.pncNumber,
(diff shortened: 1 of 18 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsintegrationapi/models/probationoffendersearch/Offender.kt
```diff
@@ -33,7 +33,7 @@ data class Offender(
           deliusCrn = otherIds.crn,
         ),
       pncId = otherIds.pncNumber,
-      hmppsId = otherIds.crn,
+      hmppsId = if (otherIds.crn?.isNotEmpty() == true) otherIds.crn else otherIds.nomsNumber,
```
CslBing, line 36: Would we want to return nonsNumber as hmppsId?  Also would probation search return person without CRN? Perharps you are looking to update POSPrisoner(person search from Prisoner serach) to return hmppsId?

mxco86 (approved):

PR #476, opened 2024-09-13
PI-2469: Added reference data endpoint

pmcphee77:


src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsintegrationapi/controllers/v1/ReferenceDataController.kt
```diff
@@ -0,0 +1,172 @@
...
+import org.springframework.web.bind.annotation.RequestMapping
+import org.springframework.web.bind.annotation.RestController
+import uk.gov.justice.digital.hmpps.hmppsintegrationapi.config.AuthorisationConfig
+import uk.gov.justice.digital.hmpps.hmppsintegrationapi.models.hmpps.Response
+import uk.gov.justice.digital.hmpps.hmppsintegrationapi.services.ReferenceDataService
+
+@Hidden
+@RestController
+@EnableConfigurationProperties(AuthorisationConfig::class)
+@RequestMapping("/v1/reference-data")
```
mxco86, line 21: Can we make the URL "/v1/hmpps/reference-data" to make it clear this is HMPPS-wide reference data.

pmcphee77: Done

mxco86 (approved):

PR #509, opened 2024-11-06
HIA-763: Skip job requiring repositiory writes for Dependabot PRs

mxco86:


.github/workflows/format-code.yml
```diff
@@ -6,6 +6,7 @@ on:
 
 jobs:
   format-code:
+    if: ${{ github.actor != 'dependabot[bot]' }}
```
marcus-bcl, line 9: For info - you don't need the `${{ ... }}` any more for certain fields.  Either way works though.
```suggestion
    if: github.actor != 'dependabot[bot]'
```

marcus-bcl (dismissed):

marcus-bcl (approved):

PR #512, opened 2024-11-11
Added endpoint to return NOMS number.

stevomcallister:


src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsintegrationapi/controllers/v1/HmppsIdController.kt
```diff
@@ -37,4 +41,34 @@ class HmppsIdController(
 
     return DataResponse(response.data)
   }
+
+  @GetMapping("{encodedNomisNumber}/nomis-number")
```
mxco86, line 45: Shouldn't this be `encodedHmppsId`?

stevomcallister: yes it should

mxco86 (approved):

PR #515, opened 2024-11-13
updated regex to handle one letter suffix  X123123

stevomcallister:


stevomcallister:
> Weird! I thought Delius only supported a single letter - no idea where the 2-letter variation came from.

it may just be the test data... all the crns in the tests are CCNNNNNN

PR #518, opened 2024-11-15
updated existing method

stevomcallister:
This looks weird but...

Sometimes the hmppsId will be a crn and sometimes it will be a noms number. 

If the crn exists then it will be the crn but if no crn exists then it will be a noms number. 

However... it is possible for the hmppsid to change without the downstream service being aware. So they should always essentially verify that the hmppsId is correct - that's what this api method does. The end client is essentially verifying that the hmppsId is the one they have and if not it will give them the correct one. 

hope that makes sense.


garethmdavies (approved):
LGTM

mxco86 (approved):

PR #528, opened 2024-11-25
API method to return the Review schedule

stevomcallister:
Initial API endpoint to provide a review schedule response in order to complete end to end for inductions and reviews for PLP. 

https://dev.integration-api.hmpps.service.justice.gov.uk/v1/persons/G3342UT/plp-review-schedule

This endpoint will return a deadline date and a description to say that it is a hardcoded response.


nathanrussell-moj-digital (approved):
LGTM 👍

PR #543, opened 2024-12-16
HMAI-52 Prison prisoners endpoint addition to Prison controller

wcdkj:
Prison prisoners endpoint addition to Prison controller to enable restricting consumer access allowing us to provide prisoner search without providing probation data.


ben-madley-mt (approved):
Happy after the changes

PR #545, opened 2024-12-18
 HMAI-88 Add filters to consumer config

wcdkj:
A filter property has been added to the consumer config data class in order to further restrict our consumer endpoint access


ben-madley-mt (dismissed):

ben-madley-mt (approved):
At some point I'd like more docs around filters but I'm happy for now. Let's merge tomorrow.

PR #557, opened 2025-01-03
HMAI-136 400 response if HMPPS ID does not match format

emmalanigan:
Whilst adding a 404 response if there was no person on prisoners endpoint, we identified we were still getting a 200 response when the HMPPS ID did not match the format expected. We now handle this as a 400 response.


ben-madley-mt (approved):
Looks great

PR #598, opened 2025-01-24
HMAI-101 Post transaction endpoint

wcdkj:
WebClient wrapper has new function to retry post based off upstream response code


src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsintegrationapi/extensions/WebClientWrapper.kt
```diff
@@ -6,13 +6,22 @@ import org.springframework.web.reactive.function.BodyInserters
...
 import uk.gov.justice.digital.hmpps.hmppsintegrationapi.models.hmpps.UpstreamApi
 import uk.gov.justice.digital.hmpps.hmppsintegrationapi.models.hmpps.UpstreamApiError
+import java.time.Duration
 
+@Suppress("ktlint:standard:property-naming")
 class WebClientWrapper(
   val baseUrl: String,
 ) {
+  val CREATE_TRANSACTION_RETRY_HTTP_CODES = listOf(500, 502, 503, 504, 522, 599, 499, 408, 301)
+  val MAX_RETRY_ATTEMPTS = 3L
```
ben-madley-mt, line 22: Why was this L required?

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsintegrationapi/prismMocks/prison-api.json
```diff
@@ -24173,7 +24173,7 @@
             "type": "string",
             "description": "Valid transaction type for the prison_id",
             "example": "CANT",
-            "enum": ["CANT,REFND,PHONE,MRPR,MTDS,DTDS,CASHD,RELA,RELS"]
+            "enum": ["CANT", "REFND", "PHONE", "MRPR", "MTDS", "DTDS", "CASHD", "RELA", "RELS"]
```
ben-madley-mt, line 24176: Can we make a PR to prison API to do this update on their side too? https://github.com/ministryofjustice/prison-api/blob/003585764f471a10048f0492725fd0040bf7442a/src/main/java/uk/gov/justice/hmpps/prison/api/model/v1/CreateTransaction.java#L27

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsintegrationapi/services/PostTransactionForPersonService.kt
```diff
@@ -0,0 +1,85 @@
...
+    if (
+      filters != null && !filters.matchesPrison(prisonId)
+    ) {
+      return Response(
+        data = null,
+        errors = listOf(UpstreamApiError(UpstreamApi.NOMIS, UpstreamApiError.Type.FORBIDDEN, "Not found")),
+      )
+    }
+
+    if (transactionRequest.type !in VALID_TRANSACTION_TYPE_CODES) {
```
ben-madley-mt, line 36: Please can we change this if and rely on the 400 error code from the API and check for bad transaction type instead, as that set up is the ultimate decider of what is allowed and won't necessarily match these types.

src/test/kotlin/uk/gov/justice/digital/hmpps/hmppsintegrationapi/mockservers/NomisApiMockServer.kt
```diff
@@ -1,9 +1,9 @@
+@file:Suppress("ktlint:standard:no-wildcard-imports")
```
ben-madley-mt, line 1: Are we able to remove this wildcard and suppression

emmalanigan:
Is there a reason we aren't handling the 409 Duplicate post - The unique_client_ref has been used before error that we can get from the API?

ben-madley-mt (changes requested):
Some qs but mainly that change in how we check transaction types

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsintegrationapi/extensions/WebClientWrapper.kt
wcdkj: variable is of desired type long so required

ben-madley-mt (approved):

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsintegrationapi/prismMocks/prison-api.json
wcdkj: looks like swagger deals with this appropriately:

allowableValues
public abstract [String](http://java.sun.com/javase/6/docs/api/java/lang/String.html?is-external=true) allowableValues
Limits the acceptable values for this parameter.
There are three ways to describe the allowable values:

To set a list of values, provide a comma-separated list. For example: first, second, third.

PR #614, opened 2025-01-29
HMAI-163 Fix issues handling upstream 400 response /v1/prison/{prisonId}/prisoners/{hmppsId}/balances 

emmalanigan:
Add BAD REQUEST error type in the web client wrapper and adjust failing tests to reflect the new behaviour.


tomwatkins1994:
Just checked some of the controllers and remembered that we'll need to handle the bad requests on those. 
PersonResponsibleOfficerController
CaseNotesController
ProtectedCharacteristicsController
and potentially any controllers using the GetPersonService, GetImageMetaDataForPersonService, GetLatestSentenceKeyDatesAndAdjustmentsForPersonService

Feels like this could be a lot of work now I've looked into it, perhaps this was the blocker stopping us from doing this before that you couldn't remember @emmalanigan ?

wcdkj (approved):

PR #616, opened 2025-02-03
Add ADR for decision around endpoint separation for Credit and Debit transactions

mattonfoot:


mattonfoot:
Having created the ADR I now feel its not useful as we are continuing with the existing solution and there is not enough information available to go either way.

PR #634, opened 2025-02-05
Address service refactor bolder

ben-madley-mt:
Created as draft as this is probably a step too far.

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsintegrationapi/services/GetAddressesForPersonService.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsintegrationapi/services/GetAddressesForPersonService.kt
@@ -37,7 +37,7 @@ class GetAddressesForPersonService(
     }
 
     if (
-      addressesFromNomis.errors.any { it.type == UpstreamApiError.Type.ENTITY_NOT_FOUND } &&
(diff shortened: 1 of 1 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsintegrationapi/services/GetAddressesForPersonService.kt
```diff
@@ -19,30 +19,40 @@ class GetAddressesForPersonService(
...
+
     if (
-      addressesFromNomis.errors.isNotEmpty() &&
-      addressesFromNomis.errors.any { it.type == UpstreamApiError.Type.ENTITY_NOT_FOUND } &&
-      addressesFromDelius.errors.isEmpty()
+      addressesFromNomis.hasError(UpstreamApiError.Type.ENTITY_NOT_FOUND) &&
+      !addressesFromDelius.hasError(UpstreamApiError.Type.ENTITY_NOT_FOUND)
     ) {
-      return addressesFromDelius
+      return Response(data = addresses, errors = errors)
```
tomwatkins1994, line 50: This line is the same as the return at the bottom of the function?

ben-madley-mt: I think I didn't push a commit.

PR #646, opened 2025-02-10
HMAI-219 Get prisoner non-associates - /v1/prison/{prisonId}/prisoners/{hmppsId}/non-associations

BushraAbdullahi:


Dockerfile.prism
```diff
@@ -23,5 +23,6 @@ RUN sed -i 's/\*\/\*/application\/json/g' /prismMocks/prisoner-offender-search.j
 
 # new files from this point on need to start with an x- followed by the port number:
 ADD https://learningandworkprogress-api-dev.hmpps.service.justice.gov.uk/v3/api-docs /prismMocks/x4020-plp-api.json
+RUN sed -i 's/\*\/\*/application\/json/g' /prismMocks/x4021-non-associations.json
```
ben-madley-mt, line 26: Why is this line needed?

wcdkj: dunno, enforce json formatting or something, we do it for all of our stored (as opposed to retrieved via network call) openapi specs that are injected into the prism container

ben-madley-mt: Ah I hadn't seen the ones above. Do you mind moving those down so all the seds are together?

tomwatkins1994 (approved):

PR #663, opened 2025-02-17
HMAI-224 - Add prison filter to /v1/hmpps/id/nomis-number/{nomisNumber}

tomwatkins1994:


tomwatkins1994:
I am wondering here if we even need to call `getPersonService.execute()` in our service or if we can just use `getPersonService.getPersonFromNomis()` since we will need to call it anyway to get the prison ID?

PR #666, opened 2025-02-18
HMAI-238 - Reduce API calls - Person and prison filter

tomwatkins1994:


ben-madley-mt (approved):
I think this isn't the final form of this change, but a step in the right direction. I don't think this should be in the get person service really, but I could be wrong.

PR #673, opened 2025-02-19
reenable test

stevomcallister:


tomwatkins1994:
It looks like the date format in the `api-docs` file now has `deadlineDate` also with the date format not in ISO format now? I can see you made a PR on the API to add an example field but I think it now has the same problem we had with `inductionPerformedAt`. 

image

PR #685, opened 2025-02-24
HMAI-223-visits-search-service-controller

wcdkj:
Search visits service and controller layer with their tests (int too).

Tagged it as Visits but not sure if we should put it under Prison instead?


src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsintegrationapi/controllers/v1/prison/PrisonController.kt
```diff
@@ -122,6 +125,50 @@ class PrisonController(
...
+    summary = "Searches for visits by prisonId and criteria.",
+    description = "<b>Applicable filters</b>: <ul><li>prisons</li></ul>",
+    responses = [
+      ApiResponse(responseCode = "200", useReturnTypeSchema = true, description = "Successfully performed the query on upstream APIs. An empty list is returned when no results are found."),
+      ApiResponse(
+        responseCode = "400",
+        description = "",
+        content = [Content(schema = Schema(ref = "#/components/schemas/BadRequest"))],
+      ),
+      ApiResponse(responseCode = "403", content = [Content(schema = Schema(ref = "#/components/schemas/ForbiddenResponse"))]),
```
emmalanigan, line 140: did you mean 403 or should this be 404?

wcdkj: added a 404 too

PR #688, opened 2025-02-25
updated to add completed at date - this is the date the user says the…

stevomcallister:
… review was completed.

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsintegrationapi/models/hmpps/ReviewSchedule.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsintegrationapi/models/hmpps/ReviewSchedule.kt
@@ -109,7 +109,7 @@ data class ReviewSchedule(
   )
   var reviewCompletedByRole: String?,
   @Schema(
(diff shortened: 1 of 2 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsintegrationapi/models/hmpps/ReviewSchedule.kt
```diff
@@ -111,7 +111,7 @@ data class ReviewSchedule(
   @Schema(
     description = "When the review schedule has a completed review this will be populated with the date of the review.",
```
nathanrussell-moj-digital, line 112: Is it worth updating the description? Something like this perhaps?

> When the review schedule has a completed review this will be populated with the date that the CIAG entered into the system when completing the Review. It is the date that the user did the Review in person, rather than the date that it was entered into the system.

And presumably we'd want to make sure the corresponding field for the Induction Schedule History has a similar description?

stevomcallister: ok updated

ben-madley-mt (approved):

PR #720, opened 2025-03-05
HMAI-89-roles-filter-implementation

wcdkj:


tomwatkins1994:
I can recreate the unit test failures locally if I am not running the local-aws-stack in docker. For some reason the unit test for GlobalConfigTest is needing this to be available?

emmalanigan (approved):

PR #736, opened 2025-03-11
HMAI-311 Add Prisons filter to /v1/persons/{encodedHmppsId}/cell-location so can be used by kilco

emmalanigan:
Add prison filter to service layer and controller

Deal with 400 bad request on the controller that can come from getNomisNumberWithPrisonFilter

Add 3 prison filter integration tests

Remove reference to encoded hmppsId


src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsintegrationapi/services/GetCellLocationForPersonService.kt
```diff
@@ -5,42 +5,41 @@ import org.springframework.stereotype.Service
...
-              errors = personResponse.errors,
-            )
-          }
-
-          personResponse.data.identifiers.nomisNumber?.let {
-            prisonerOffenderSearchGateway.getPrisonOffender(nomsNumber = it)
-          }
-        }
-      }
+    val nomisNumber =
```
tomwatkins1994, line 26: Could we get a service test to cover when `getPersonService.getNomisNumberWithPrisonFilter()` returns a NULL `NomisNumber`?

tomwatkins1994 (changes requested):

tomwatkins1994 (approved):

PR #750, opened 2025-03-13
Create a PR template that includes tests for filters

BushraAbdullahi:


.github/filters_pull_request_template.md
```diff
@@ -0,0 +1,45 @@
...
+- [ ] Added logic to throw a validation exception when a bad request error is received from the upstream service.
+- [ ] Added `ApiResponse` for HTTP 400 (Bad Request):
+  - [ ] Response code: "400".
+  - [ ] Description: "Malformed hmppsId."
+  - [ ] Content: `Content(schema = Schema(ref = "#/components/schemas/BadRequest"))`.
+
+**Controller Tests:**
+
+- [ ] Updated `hmppsId` variable to a non-encoded version (e.g., "A1234AA").
+- [ ] Modified mock to use `getPersonService.getPersonWithPrisonFilter(hmppsId = hmppsId, filters = filters)` instead of `getPersonService.execute`.
```
tomwatkins1994, line 24: `getPersonService.getPersonWithPrisonFilter()` usually wouldn't be called in the controller unless the endpoint was only getting `Person` data

.github/filters_pull_request_template.md
```diff
@@ -0,0 +1,45 @@
...
+- [ ] Updated service call to include `filters`.
+- [ ] Added a test case for the bad request error.
+
+**Service Logic:**
+
+- [ ] Added `filters` as a parameter to the service function.
```
tomwatkins1994, line 30: Maybe add here that it should currently be nullable

.github/filters_pull_request_template.md
```diff
@@ -0,0 +1,45 @@
...
+- [ ] Updated the service call to use `getPersonService.getPersonWithPrisonFilter(hmppsId = hmppsId, filters = filters)`.
```
tomwatkins1994, line 31: We may also be using `getPersonService.getNomisNumberWithPrisonFilter()` if we only need the NOMIS number in the service.

.github/filters_pull_request_template.md
```diff
@@ -0,0 +1,45 @@
...
+
+**Service Tests:**
+
+- [ ] Updated `hmppsId` variable to a non-encoded version (e.g., "A1234AA").
+- [ ] Modified mock to use `getPersonService.getPersonWithPrisonFilter(hmppsId = hmppsId, filters = filters)` instead of `getPersonService.execute`.
+
+**Integration Tests:**
+
+- [ ] Updated `globals.yml` with the modified endpoint.
```
tomwatkins1994, line 40: Specifically we should highlight that we are adding it to the `private-prisons` role.include

tomwatkins1994 (changes requested):

tomwatkins1994 (approved):

PR #763, opened 2025-03-20
HIA-775 Implement LAO Redaction

anthony-britton-moj:


build.gradle.kts
```diff
@@ -9,11 +11,13 @@ configurations {
 
 dependencies {
   runtimeOnly("org.flywaydb:flyway-database-postgresql")
-  implementation("org.springframework.boot:spring-boot-starter-webflux:3.4.3")
+  implementation("org.springframework.boot:spring-boot-starter-webflux")
```
ben-madley-mt, line 14: Do you know what has happened with the removal of versions in this file?

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsintegrationapi/limitedaccess/redactor/LaoRedactorAspect.kt
SteveLeachMoj, line 1: Ideally I'd like to get a better understanding of why we need the aspect and can't do this in the Interceptor. My understanding is that the ModelAndView parameter in the interceptor call is null, so we can't redact the model, but it would be good to know _why_ it is null and whether there is anything we can do about that.

SteveLeachMoj: Summary of Slack discussion...
- The endpoints are implemented using `@RestController` which is just `@Controller` and `@ResponseBody`
- `@ResponseBody` instructs the framework to directly render the response as JSON, so there is no view to render the model
- The asymmetry between LAO lookup (in the Interceptor) and redaction (in AoP) could be confusing, but is not a showstopper
- We could consider switching to `ResponseBodyAdvice` implementations as a follow-up PR

SteveLeachMoj: I like this change @anthony-britton-moj 

You arguably are still following the Single Responsibility principle, because the `LaoRedactorAspect` delegates the identification bit to the `laoChecker` which is injected in. 

I suppose you could potentially make it cleaner still by having `laoChecker` accept the `hmppsId` as the parameter
`crnSupplier.getCrn(hmppsId)?.let { loaChecker.getAccessForCrn(it)?.asLaoContext() }`

SteveLeachMoj: Maybe, but I'd probably call putting the LAO Access Context in the request context an optimisation rather than anything to do with single responsibility patterns; anything that needs LAO context can ask for it. Explicit requests for that information do have some code maintainability benefits over passing it around in a context.

I think it will also simplify if/when we move to a `ResponseBodyAdvice` implementation as we'd then be working directly with Spring MVC objects (HttpRequest and typed response body objects), but that's for later.

SteveLeachMoj: Sorry, I should have been clearer. 

The Aspect code has no inherent understanding of the objects it is working with, so you need to make assumptions or do type checks & casts such as 

`val hmppsId = (joinPoint.args.first() as String)`

There's no inherent understanding that this is an HTTP URL parameter.

On the response object side, you have 

`when (result) {
        is DataResponse<*> -> redactDataResponse(result)`

as opposed to `ResponseBodyAdvice<DataResponse<DynamicRisk>>` for example.

This isn't a big problem, but I suspect that the code would be slightly cleaner with ResponseBodyAdvice.

PR #821, opened 2025-04-11
HMAI-443 /v1/persons/{hmppsId}/visit-orders should not return 404 if 0 values found

BushraAbdullahi:


src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsintegrationapi/services/GetVisitOrdersForPersonService.kt
tomwatkins1994, line 1: Looks good, couple of things I think we could add to improve this:

1. The type of `transformedResponse` could probably be `Response<VisitOrders>` as we don't ever set it to be `null`
2. Could we check for errors on `visitBalancesResponse` and return if there are any straight after doing the lookup instead of at the end of the file? So on line 33
3. The null check would then be `visitBalancesResponse.data` instead of `transformedResponse.data`

src/test/kotlin/uk/gov/justice/digital/hmpps/hmppsintegrationapi/services/GetVisitOrdersForPersonServiceTest.kt
```diff
@@ -109,4 +109,26 @@ class GetVisitOrdersForPersonServiceTest(
...
         ).shouldBe(true)
     }
+
+    it("return a 404 error when getVisitBalances data is null") {
+      whenever(nomisGateway.getVisitBalances(nomisNumber)).thenReturn(
+        Response(
+          data = null,
+          errors =
+            listOf(
+              UpstreamApiError(
```
tomwatkins1994, line 119: Since we check for null and return the error in the check we don't need to return an error in this mock

tomwatkins1994 (approved):

PR #825, opened 2025-04-14
HMAI-432 Get Number of Children - /v1/persons/number-of-children Gateway

BushraAbdullahi:
- Gateway function to query number of children for a prisoner from the personal relationships api
- unit test coverage


src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsintegrationapi/models/personalRelationships/NumberOfChildren.kt
tomwatkins1994, line 1: Could prefix this with PR like we have on the other models? Like `PRPrisonerContact` for example

tomwatkins1994 (changes requested):
Looks good, just a minor point on the naming of the model

tomwatkins1994 (approved):

PR #849, opened 2025-04-16
HMAI-396 - Fix  /v1/person/{hmppsId}/person-responsible-officer 404s

tomwatkins1994:
This PR makes changes to `/v1/person/{hmppsId}/person-responsible-officer` to prevent an issue with us getting 404s when calling this endpoint.

* If we get a 404 from either the POM or nDelius gateway we ignore it and just return null in the respective fields, all other errors are still treated as errors though
* We still check the person exists and return 404 if not (including by checking filters)
* The logic being that the person may not be in delius and therefore not have a community manager
* Manage POM gateway returns this error message on 404, which doesn't feel like something we should be treating as an error on our side 
```json
{
    "status": "error",
    "message": "Not ready for allocation"
}
```
* I also noticed some obvious tidy ups and some missing test coverage
* TODO: When checking this endpoint in dev, we may get a 401 due on the nDelius endpoint, I could not access it from postman with our creds and this has masked until now by the 404 from Manage POM


ben-madley-mt:
Responsible Officer should be indicating who is _the_ responsible officer. This should probably be tackled as a second change.

ben-madley-mt (approved):

PR #851, opened 2025-04-17
HMAI-470 Remove Feature flag from /v1/persons/{encodedHmppsId}/images

emmalanigan:
Following the implementation of new functionality with prison filters behind a feature flag, we have tested in dev and are happy that the endpoint is still functioning as expected. 

This ticket therefore, removes the feature flag.


ben-madley-mt (approved):
Looks good. Good reminder we need to work out a way to document the tests we run in dev.

PR #854, opened 2025-04-22
Feature flag impl, number-of-children

wcdkj:


src/test/kotlin/uk/gov/justice/digital/hmpps/hmppsintegrationapi/extensions/featureflag/FeatureFlagAspectTest.kt
```diff
@@ -0,0 +1,40 @@
...
+import org.mockito.kotlin.verify
+import uk.gov.justice.digital.hmpps.hmppsintegrationapi.config.FeatureFlagConfig
+import uk.gov.justice.digital.hmpps.hmppsintegrationapi.exception.FeatureNotEnabledException
+
+class FeatureFlagAspectTest {
+  private var proceedingJoinPoint: ProceedingJoinPoint =
+    mock(ProceedingJoinPoint::class.java).apply {
+      `when`(proceed()).thenReturn(Unit)
+    }
+  private val featureFlagConfig: FeatureFlagConfig = FeatureFlagConfig(useArnsEndpoints = true, useNumberOfChildrenEndpoints = false)
```
tomwatkins1994, line 19: Could we rename the flags to ones not used in the codebase? I initially thought we were testing against actual flags before looking further. Could call them something generic like `falseFlag` and `trueFlag`?

tomwatkins1994 (changes requested):

src/test/kotlin/uk/gov/justice/digital/hmpps/hmppsintegrationapi/extensions/featureflag/FeatureFlagAspectTest.kt
wcdkj: the fields in the config object? they're static so nope

tomwatkins1994: Only worry is that when those fields get removed in a later PR, we'll need to change the tests to use something else. Could we not just add them to the config?

wcdkj: not a fan of modifying models with inert fields to ensure tests pass
we could create a delegate but at that point it feels overkill for what we're doing and then we're not touching exact implementation models
i think it's more reasonable to just have a test fail if we remove an expected field

tomwatkins1994 (approved):

PR #857, opened 2025-04-23
Add visits DLQ secret

tomwatkins1994:
This PR adds the secret for the visit writes DLQ to our env variables via helm so we have the ability to manage it via our queue endpoints (the automatically generated ones).


tomwatkins1994:
Closing as we shouldn't have any management of the DLQ

PR #879, opened 2025-04-29
HMAI-487  Get location information - /v1/prison/{prisonId}/location/{key} - Controller Layer

BushraAbdullahi:
- write controller function
- add controller tests
- add feature flag
- add integration tests

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsintegrationapi/services/GetLocationByKeyService.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsintegrationapi/services/GetLocationByKeyService.kt
@@ -12,25 +12,25 @@ import uk.gov.justice.digital.hmpps.hmppsintegrationapi.models.roleconfig.Consum
 class GetLocationByKeyService(
   @Autowired val locationsInsidePrisonGateway: LocationsInsidePrisonGateway,
   @Autowired val consumerPrisonAccessService: ConsumerPrisonAccessService,
-)
+) {
+  fun execute(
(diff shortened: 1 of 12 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsintegrationapi/services/GetLocationByKeyService.kt
```diff
@@ -0,0 +1,40 @@
...
+    filters: ConsumerFilters?,
+  ): Response<Location?> {
+    val checkAccess = consumerPrisonAccessService.checkConsumerHasPrisonAccess<Location?>(prisonId, filters)
+
+    if (checkAccess.errors.isNotEmpty()) {
+      return Response(data = checkAccess.data, errors = checkAccess.errors)
+    }
+
+    if (!key.startsWith("$prisonId-")) {
+      throw ValidationException("Key must start with matching prisonId")
```
tomwatkins1994, line 29: Any reason why we're throwing and not just adding to the Response errors?

scripts/smoke-tests.sh
```diff
@@ -43,6 +43,7 @@ visitReference="qd-lh-gy-lx"
 clientReference="123456"
 contactId="1898610"
 imageId="1988315"
+locationIdKey="A"
```
tomwatkins1994, line 46: Should be MKI-A

tomwatkins1994 (changes requested):

tomwatkins1994 (approved):

PR #903, opened 2025-05-06
HMAI-521 - Replace existing api docs with swagger docs

tomwatkins1994:
Due to multiple issues with using the API docs generated by the `tech-docs-gem` it has been decided that we should now use the Swagger docs in their place. This PR embeds the swagger docs in the existing page so existing links will still work.

This is how it now looks:

image

It also shows this message if JavaScript is not enabled:

image


SteveLeachMoj (approved):
I'm happy to merge this

PR #928, opened 2025-05-16
HMAI-545 - Planet fm ref renamed to external ref

tomwatkins1994:
`planetFmReference` has been renamed to `externalReference` so we don't expose the name of another supplier outside of HMPPS. This is also now included in `DeactivateLocationRequest` but is not sent as part of the request to the location service for now.


src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsintegrationapi/models/hmpps/DeactivateLocationRequest.kt
```diff
@@ -17,9 +17,8 @@ data class DeactivateLocationRequest(
   val deactivationReasonDescription: String? = null,
   @Schema(description = "Estimated reactivation date", example = "2025-01-05", required = false)
   val proposedReactivationDate: LocalDate? = null,
-  @Schema(description = "Planet FM reference number", example = "23423TH/5", required = false)
-  @field:Size(max = 60, message = "Planet FM reference number cannot be more than 60 characters")
```
emmalanigan, line 21: do we not still need this validation? or it's not being passed through anymore so not needed?

emmalanigan (approved):

PR #948, opened 2025-06-06
HMAI-507 - Remove v1/config/authorisation endpoint

tomwatkins1994:
Integration events should now be using `v2/config/authorisation`.


emmalanigan (approved):

tomwatkins1994:
I have checked the logs in prod and can see that the v1 endpoint is no longer being called and v2 is being called in its place

PR #950, opened 2025-06-09
HMAI-599 Edit events consumer setup documentation

emmalanigan:
Whilst completing some manual testing on our events filtering work in the hmpps-integration-events repository, we identified changes that needed to be made to the process of setting up a new consumer on the events queue.

We identified the terraform with the aws_secretsmanager_secret and aws_secretsmanager_secret_version, SNS topic subscription needs to be done after the terraform that creates the secret has run, so that you can manually add the secret name. 

The documentation has been updated to reflect that the new structure should be:

- [secret.tf](http://secret.tf/) PR
- Manually initialise secret with default
- Do a second PR for the rest

Make it clear that initialising the filter policy secret is required (and should happen in between the two terraform steps above)


ben-madley-mt (changes requested):
Can we move steps 2 and 3 in your version to after step 6. Get all of the secrets stuff done and then do the remainder of the setup.

BushraAbdullahi (approved):

PR #958, opened 2025-06-10
HMAI-391 - Update middleman

tomwatkins1994:


tomwatkins1994:
Looks like the tech-docs-gem is actually pinning middleman to 4.5.1 due to a regression in 4.6.0. So we probably don’t want to do do this

[undefined method 'app' for nil (NoMethodError) - middleman sitemap error · Issue #2818 · middleman/middleman](https://github.com/middleman/middleman/issues/2818)

PR #971, opened 2025-06-12
HMAI-620 Get Prison Activities – GET /v1/prison/{prisonId}/activities

emmalanigan:


src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsintegrationapi/models/activities/ActivitiesActivityCategory.kt
```diff
@@ -1,8 +1,18 @@
...
 data class ActivitiesActivityCategory(
   val id: Long,
   val code: String,
   val name: String,
   val description: String?,
-)
+) {
+  fun toActivityCategory() =
+    ActivityCategory(
+      id = this.id,
```
tomwatkins1994, line 13: Do we need the internal ID?

PR #984, opened 2025-06-18
Randomly set certificate serial number

ben-madley-mt:
This is instead of using `01` for all certificates which contravenes the spec and makes it harder to invalidate certificates.


SteveLeachMoj (approved):
https://dsdmoj.atlassian.net/browse/HIA-793

PR #1005, opened 2025-06-24
HMAI-642 Get prisoner’s scheduled instances – GET /prisons/{prisonId}/prisoners/{hmppsId}/scheduled-instances

BushraAbdullahi:
- Service Layer & Tests
- Controller Layer & Tests
- Feature Flag (false for pre-prod and prod) & Feature Flag Integration Test
- Integration Tests
- Smoke Test


tomwatkins1994 (commented):
Could we change the URL of this endpoint to `/v1/prison/{prisonId}/prisoners/{hmppsId}/scheduled-instances`?

PR #1007, opened 2025-06-25
HMAI-660 - GET De-allocation reasons – v1/activities/deallocation-reasons

BushraAbdullahi:
- Service Layer & Tests
- Controller Layer & Tests
- Feature Flag (false for pre-prod and prod) & Feature Flag Integration Test
- Integration Tests
- Smoke Test


ben-madley-mt (changes requested):
Can we talk about tags

PR #1052, opened 2025-07-09
PI-3252: Create api filter feature flag

pmcphee77:
This PR will (behind a feature flag) pass the list of alert codes to the alerts api for filtering the PND alert code to be performed within the api.
e.g `https://alerts-api-preprod.hmpps.service.justice.gov.uk/prisoners/<id>/alerts?page=0&size=10&alertCode=BECTER,HA,XA,XCA,XEL,XELH,XER,XHT,XILLENT,XIS,XR,XRF,XSA,HA2,RCS,RDV,RKC,RPB,RPC,RSS,RST,RDP,REG,RLG,ROP,RRV,RTP,RYP,HS,SC`
This will keep the pagination intact and will reduce required number of calls from the client from 5 to 1 (in the example above)
If the feature flag is not enabled, then existing processing will be followed and the pagination for PND will continue to be broken
For non pnd requests the requests will continue as normal (without the codes in the query param)
FYI - i have tested the above URL in preprod and the filters work

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsintegrationapi/controllers/v1/person/AlertsController.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsintegrationapi/controllers/v1/person/AlertsController.kt
@@ -15,10 +15,13 @@ import org.springframework.web.bind.annotation.RequestAttribute
 import org.springframework.web.bind.annotation.RequestMapping
 import org.springframework.web.bind.annotation.RequestParam
 import org.springframework.web.bind.annotation.RestController
+import uk.gov.justice.digital.hmpps.hmppsintegrationapi.config.FeatureFlagConfig
+import uk.gov.justice.digital.hmpps.hmppsintegrationapi.config.FeatureFlagConfig.Companion.USE_ALERTS_API_FILTER
 import uk.gov.justice.digital.hmpps.hmppsintegrationapi.exception.EntityNotFoundException
 import uk.gov.justice.digital.hmpps.hmppsintegrationapi.models.hmpps.Alert
 import uk.gov.justice.digital.hmpps.hmppsintegrationapi.models.hmpps.UpstreamApiError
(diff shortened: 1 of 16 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsintegrationapi/controllers/v1/person/AlertsController.kt
```diff
@@ -83,7 +87,7 @@ class AlertsController(
     @Parameter(description = "The maximum number of results for a page", schema = Schema(minimum = "1")) @RequestParam(required = false, defaultValue = "10", name = "perPage") perPage: Int,
     @RequestAttribute filters: ConsumerFilters?,
   ): PaginatedResponse<Alert> {
-    val response = getAlertsForPersonService.execute(hmppsId, filters, page, perPage, pndOnly = true)
+    val response = if (featureFlagConfig.isEnabled(USE_ALERTS_API_FILTER)) getAlertsForPersonService.getAlerts(hmppsId, filters, page, perPage, PAPaginatedAlerts.PND_ALERT_CODES) else getAlertsForPersonService.execute(hmppsId, filters, page, perPage, pndOnly = true)
```
SteveLeachMoj, line 90: Do we need to (or even should we?) change this deprecated endpoint?

There's a reasonable case that the "fixed" behaviour is only needed on the non-deprecated version.

pmcphee77: Agreed - changed!

tomwatkins1994 (changes requested):
Looks good but I wonder if we should just add an integration test for the new codepath? See `AlertsIntegrationTest`

pmcphee77:
> Looks good but I think we should add some integration tests for the new codepath. See `AlertsIntegrationTest`

Ive added a separate integration test `AlertsWithApiQueryFeatureIntegrationTest` with the feature turned on

PR #1056, opened 2025-07-10
HMAI-694 Get activity schedule suitability criteria - AMEND MODELS

emmalanigan:
In the creation of the endpoint in the Activities API, we identified:

- Pay rates were actually returned as a list
- Pay rate includes a start date field
- PrisonPayBand is returned with more data so should use the ActivitiesPrisonPayBand model at the gateway level, before being transformed.

Tests have also been updated to reflect these changes.

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsintegrationapi/models/activities/ActivitiesPayRate.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsintegrationapi/models/activities/ActivitiesPayRate.kt
@@ -21,5 +21,6 @@ data class ActivitiesPayRate(
       rate = this.rate,
       pieceRate = this.pieceRate,
       pieceRateItems = this.pieceRateItems,
+      startDate = this.startDate,
     )
 }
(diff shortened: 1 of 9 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsintegrationapi/models/activities/ActivitiesPayRate.kt
```diff
@@ -1,22 +1,23 @@
...
   val incentiveLevel: String,
-  val prisonPayBand: PrisonPayBand,
-  val rate: Int,
-  val pieceRate: Int,
-  val pieceRateItems: Int,
+  val prisonPayBand: ActivitiesPrisonPayBand,
+  val rate: Int?,
+  val pieceRate: Int?,
+  val pieceRateItems: Int?,
+  val startDate: LocalDate?,
```
tomwatkins1994, line 14: Is start date just a field that got missed? Let's copy it into our hmpps model if so.

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsintegrationapi/models/hmpps/SuitabilityCriteria.kt
```diff
@@ -8,7 +8,7 @@ data class SuitabilityCriteria(
   @Schema(description = "Whether the activity is a paid activity", example = "true")
   val isPaid: Boolean,
   @Schema(description = "The pay rate by incentive level and pay band that can apply to this activity")
-  val payRate: PayRate,
+  val payRate: List<PayRate>,
```
tomwatkins1994, line 11: With this now being a list, let's rename it to `payRates` (and obviously we will need to adjust this on the activities API too)

tomwatkins1994 (approved):

PR #1058, opened 2025-07-10
Test and update feature flags for GET /prison/{prisonId}/prisoners/{hmppsId}/scheduled-instances

BushraAbdullahi:
Returns a 200 response:
image

Returns a 400 when a bad request is made:
image

Returns a 404 when the hmpps Id cannot be found:
image


tomwatkins1994:
> I have managed to find a combination for prison/prison number which returns data - looks like we have some issues on the endpoint:
> 
> Prison ID: RSI HMPPS ID: G6933GQ
> 
> image

Have retested following fixes and is now returning data as expected

tomwatkins1994 (approved):

PR #1063, opened 2025-07-11
HIA-777: Throw exception if lao context cannot not be retrieved

pmcphee77:


src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsintegrationapi/config/FeatureFlagConfig.kt
```diff
@@ -34,6 +34,7 @@ data class FeatureFlagConfig(
     const val USE_SUITABILITY_ENDPOINT = "use-suitability-endpoint"
     const val USE_HISTORICAL_ATTENDANCES_ENDPOINT = "use-historical-attendances-endpoint"
     const val USE_WAITING_LIST_ENDPOINT = "use-waiting-list-endpoint"
+    const val USE_NO_LAO_CONTEXT_EXCEPTION = "use-no-lao-context-exception"
```
SteveLeachMoj, line 37: We don't need to start all our feature flag names with "USE_", that is just a habit we have fallen into. Maybe `ERROR_ON_NO_LAO_CONTEXT` or something?

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsintegrationapi/config/HmppsIntegrationApiExceptionHandler.kt
```diff
@@ -215,6 +216,20 @@ class HmppsIntegrationApiExceptionHandler {
       )
   }
 
+  @ExceptionHandler(LimitedAccessFailedException::class)
```
SteveLeachMoj, line 219: Why do we need a specific error handler for this new exception? Doesn't the standard one for `RuntimeException` handle it well?

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsintegrationapi/config/FeatureFlagConfig.kt
pmcphee77: Ok - ive made the change

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsintegrationapi/config/HmppsIntegrationApiExceptionHandler.kt
SteveLeachMoj: Fair enough. It seems like there is something we can do generically to improve this pattern, but I'll create a separate ticket for it.

PR #1065, opened 2025-07-11
HIA-778: Create basic status endpoint

pmcphee77:


pmcphee77:
> Could we get a smoke test (in `full-access-smoke-test.js`) and a simple integration test for this endpoint?

Ive added `src/test/kotlin/uk/gov/justice/digital/hmpps/hmppsintegrationapi/integration/StatusIntegrationTest.kt`

tomwatkins1994 (approved):

PR #1120, opened 2025-07-22
HIA-801: Remove arns feature flag

pmcphee77:


src/test/kotlin/uk/gov/justice/digital/hmpps/hmppsintegrationapi/extensions/featureflag/FeatureFlagAspectTest.kt
```diff
@@ -19,8 +21,7 @@ class FeatureFlagAspectTest {
   private val featureFlagConfig: FeatureFlagConfig =
     FeatureFlagConfig(
       mapOf(
-        "use-arns-endpoints" to true,
-        "use-education-assessments-endpoints" to false,
+        "use-education-assessments-endpoints" to true,
```
SteveLeachMoj, line 24: Can we use a synthetic key name here? We don't need a constant for it, as it is just for this test.

SteveLeachMoj (approved):

PR #1128, opened 2025-07-30
HI-837: Prisoner Search API Discriminator validation fix

pmcphee77:
Ticket [HIA-837](https://dsdmoj.atlassian.net/browse/HIA-837) raised the issue where the following error is encountered whilst validating the prisoner search Open API specification in the tests

`"'discriminator' field 'type' must be defined as a string property"`

Background

The prisoner search uses a 'discriminator' to support polymorphism in the API definition (i.e a shared Matcher definition for BooleanMatcher, StringMatcher, IntMatcher etc).

It is currently defined like this:

```
      "Matcher": {
        "type": "object",
        "discriminator": {
          "propertyName": "type",
        },
        "properties": {
          "type": {
            "type": "string"
          }
        },
        "required": ["type"]
      }
```

The discriminator section is therefore validated by the `com.atlassian.oai.validator.schema.keyword.DiscriminaterSyntaxChecker` which uses Swagger Core 2.X

Problem

Now why would we be getting the error "'discriminator' field 'type' must be defined as a string property" when we can clearly see that the discriminator 'type' property has its type set to 'string'.

Well, as we can see in the prisoner search api specification, the version of the schema is `"openapi": "3.1.0"`.

One of the differences between 3.0 and 3.1 is that version 3.1 allows for an array to be specified for the type.

After investigation it appears that there was a change in Swagger Core 2.X, when processing OpenAPI 3.1 specifications, the type field is mapped by Set<String> types member (instead of String type) to also support array data type.

Therefore the discriminator is deserialized into a Schema class with populated types member (while type remains null) and looks something like this.

```
    discriminator: {
       type: null,
       types: ["string"]
    }
```
Therefore whilst validating using `com.atlassian.oai.validator.schema.keyword.DiscriminaterSyntaxChecker` the error is more obvious.
`"'discriminator' field 'type' must be defined as a string property"`

Fix(es)

How do we fix this? Well the swagger wiki [Swagger-2.X---OpenAPI-3.1](https://github.com/swagger-api/swagger-core/wiki/Swagger-2.X---OpenAPI-3.1#foreword) describes a workaround that has been implemented for this very issue to allow for a non breaking change.
If you set the system property bind-type=true, then this does not bind single value types to an array and leaves the functionality intact.

Therefore i have created a BindTypeValidationListener class that sets this flag around the requestReceived method of the com.atlassian.oai.validator.OpenApiInteractionValidator class (depending on whether it is an openApi V3.1.0 schema. This is used when creating each of the MockServers used in the tests and 

This fixes the above error.

However, this then unearthed a stream of other errors.

The reason for this is that the prisoner search Open API schema doesnt actually provide a mechanism for matching the discriminator value it uses in the request e.g ("String", "Int", "PNC" etc) and is also incorrect in places - for example using one.
Therefore, to get a fully working and validate-able schema, i needed to make the following changes.

1. Provide a mapping between the discriminator value and the Matcher schema

```
          "mapping": {
            "String": "StringMatcher",
            "Boolean": "BooleanMatcher",
            "Date": "DateMatcher",
            "DateTime": "DateTimeMatcher",
            "Int": "IntMatcher",
            "PNC": "PncMatcher"
          }
```
2. For the list of matchers in the query schema. Replace the oneOf block with the list of child classes with a single reference to Matcher parent class. Using a oneOf causes cyclic reference issues which causes complicated loops.

4. In each of the matchers provide an enum val to enforce that the matcher must have the discriminator value equal to the value of the enum. This will help any oneOf matchers to match to the correct child. It specifies that it must be the value in the description, so better to enforce it.
```
                "type": "string",
                "description": "Must be String",
                "enum": ["String"],
```
5. Finally remove the allOf references from the child back to the parent as there would be a cyclic reference causing infinite loops.

I think this may need to be passed to the prisons team to see if they will implement these changes

[HIA-837]: https://dsdmoj.atlassian.net/browse/HIA-837?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/test/kotlin/uk/gov/justice/digital/hmpps/hmppsintegrationapi/integration/prison/DeactivateLocationIntegrationTest.kt
+++ b/src/test/kotlin/uk/gov/justice/digital/hmpps/hmppsintegrationapi/integration/prison/DeactivateLocationIntegrationTest.kt
@@ -1,5 +1,6 @@
 package uk.gov.justice.digital.hmpps.hmppsintegrationapi.integration.prison
 
+import com.atlassian.oai.validator.wiremock.OpenApiValidationListener
 import com.fasterxml.jackson.module.kotlin.jacksonObjectMapper
 import io.kotest.assertions.json.shouldContainJsonKeyValue
 import io.kotest.matchers.shouldBe
@@ -8,8 +9,13 @@ import org.awaitility.kotlin.matches
 import org.awaitility.kotlin.untilCallTo
 import org.junit.jupiter.api.AfterEach
 import org.junit.jupiter.api.Test
+import org.junit.jupiter.api.assertThrows
+import org.springframework.beans.factory.annotation.Autowired
+import org.springframework.http.HttpStatus
 import org.springframework.test.web.servlet.result.MockMvcResultMatchers.content
 import org.springframework.test.web.servlet.result.MockMvcResultMatchers.status
+import org.springframework.web.reactive.function.client.WebClientResponseException
+import uk.gov.justice.digital.hmpps.hmppsintegrationapi.gateways.PrisonerOffenderSearchGateway
 import uk.gov.justice.digital.hmpps.hmppsintegrationapi.integration.IntegrationTestWithQueueBase
 import uk.gov.justice.digital.hmpps.hmppsintegrationapi.models.hmpps.DeactivateLocationRequest
 import uk.gov.justice.digital.hmpps.hmppsintegrationapi.models.hmpps.DeactivationReason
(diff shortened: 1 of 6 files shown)


src/test/resources/openapi-specs/prisoner-search.json
SteveLeachMoj, line 1: Is this the latest version of the upstream spec? Assuming it is, would it be possible to break it out into its own PR, with a description saying that it is an upstream spec refresh, then rebase this PR onto that one?
This would more clearly distinguish between upstream changes (the updated spec) and our own changes (which depend on that spec update)?

src/test/kotlin/uk/gov/justice/digital/hmpps/hmppsintegrationapi/integration/prison/DeactivateLocationIntegrationTest.kt
SteveLeachMoj, line 1: Could we have a negative test as well, with a json file that doesn't match the schema?

SteveLeachMoj (changes requested):

SteveLeachMoj (approved):

PR #1159, opened 2025-08-29
MAP-2341 Add CaseNotes filter

danbenton-mojdt:
To restrict access to defined top-level Case Notes types

Also provides geoamey client in dev access to Case Notes endpoint with limited types


dafhoustonmoj (approved):
LGTM

PR #1195, opened 2025-09-15
Hia 820 role defs to dsl

pmcphee77:


src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsintegrationapi/roles/PrisonerEscortCustodyServiceRole.kt
```diff
@@ -0,0 +1,20 @@
+package uk.gov.justice.digital.hmpps.hmppsintegrationapi.roles
+
+import uk.gov.justice.digital.hmpps.hmppsintegrationapi.roles.dsl.role
+
+val prisonerEscortCustodyServiceRole =
```
SteveLeachMoj, line 5: Very nice DSL implementation

SteveLeachMoj:
Would it be worth adding some unit tests for the roles?

For example, we could have a test that `FullAccessRole.includes` is the same as `AllEndpointsRole.includes`. 

Other potential invariants we could test:-

- `ReferenceDataOnlyRole.includes` has only 2 entries
- `PrivatePrisonRole.filters.prisons` is not null
- `FullAccessRole` has no filters

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsintegrationapi/extensions/AuthorisationFilter.kt
```diff
@@ -72,7 +76,11 @@ class AuthorisationFilter(
     val rolesInclude =
       buildList {
         for (consumerRole in consumersRoles.orEmpty()) {
-          addAll(globalsConfig.roles[consumerRole]?.include.orEmpty())
+          if (featureFlagConfig.isEnabled(USE_ROLES_DSL)) {
+            addAll(getRoles()[consumerRole]?.include.orEmpty())
```
SteveLeachMoj, line 80: Could we make `roles` an `internal` mutable field of `AuthorisationFilter`, initialised to `RoleMapping.roles`? That would be everything needed for production use, but unit tests could just modify it as needed and we wouldn't need to do any mocking.

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsintegrationapi/roles/AllEndpointsRole.kt
```diff
@@ -0,0 +1,76 @@
+package uk.gov.justice.digital.hmpps.hmppsintegrationapi.roles
+
+import uk.gov.justice.digital.hmpps.hmppsintegrationapi.roles.dsl.role
+
+val allEndpointsRole =
+  role {
+    include {
+      // Provides access to all endpoints
```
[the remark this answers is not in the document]
pmcphee77: Yes - ive created a constants that can be defined in the same format as the DSL e.g.

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsintegrationapi/extensions/AuthorisationFilter.kt
```diff
@@ -24,6 +27,7 @@ import java.io.IOException
 class AuthorisationFilter(
   @Autowired val authorisationConfig: AuthorisationConfig,
   @Autowired val globalsConfig: GlobalsConfig,
+  @Autowired val featureFlagConfig: FeatureFlagConfig,
```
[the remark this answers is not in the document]
pmcphee77: Ok, ive done it this way - but this will probably be changed in the next iteration as i will need to inject a differerent component here which will require a component to be injected - to get mappa detail

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsintegrationapi/extensions/AuthorisationFilter.kt
SteveLeachMoj: Could you expand on this? We'll just be adding the mappa filters alongside the existing prison and case note filters, so the contact events endpoint code will get it automatically.

pmcphee77:
> Would it be worth adding some unit tests for the roles?
> 
> For example, we could have a test that `FullAccessRole.includes` is the same as `AllEndpointsRole.includes`.
> 
> Other potential invariants we could test:-
> 
> * `ReferenceDataOnlyRole.includes` has only 2 entries
> * `PrivatePrisonRole.filters.prisons` is not null
> * `FullAccessRole` has no filters

Added tests - FYI - PrivatePrisonRole.filters.prisons should be null based on globals yaml, also FullAccessRole - does have filters

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsintegrationapi/extensions/AuthorisationFilter.kt
pmcphee77: Changed

pmcphee77:
> > Would it be worth adding some unit tests for the roles?
> > For example, we could have a test that `FullAccessRole.includes` is the same as `AllEndpointsRole.includes`.
> > Other potential invariants we could test:-
> > 
> > * `ReferenceDataOnlyRole.includes` has only 2 entries
> > * `PrivatePrisonRole.filters.prisons` is not null
> > * `FullAccessRole` has no filters
> 
> Added tests - FYI - PrivatePrisonRole.filters.prisons should be null based on globals yaml, also FullAccessRole - does have filters

Added FiltersExtractionFilterRoleDslTest that compares the filters created for each of the roles - both from DSL and GlobalConfig

SteveLeachMoj (dismissed):

PR #1209, opened 2025-09-25
HIA-820: Remove dsl roles feature flag and remove globals config

pmcphee77:


src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsintegrationapi/controllers/v2/ConfigController.kt
```diff
@@ -36,17 +31,13 @@ class ConfigController(
...
-          addAll(globalsConfig.roles[consumerRole]?.include.orEmpty())
-        }
+        addAll(roles[consumerRole]?.include.orEmpty())
       }
       addAll(consumerConfig?.include.orEmpty())
     }
 
   private fun buildFiltersList(consumerConfig: ConsumerConfig?): ConsumerFilters? {
-    val aggregatedRoles: List<Role>? = consumerConfig?.roles?.mapNotNull { globalsConfig.roles[it] }
+    val aggregatedRoles: List<Role>? = consumerConfig?.roles?.mapNotNull { roles[it] }
```
pmcphee77, line 40: Ive noticed that this was not previously feature flagged, but this is for the config controller only which i dont think is widely used. 
I could park this PR and feature flag this before i continue. But i think it will be ok

SteveLeachMoj: The config endpoint is for internal use only, so I'm relatively comfortable with it.

SteveLeachMoj (approved):

PR #1400, opened 2025-12-04
HIA-1139 POC for API Spec republishing

SteveLeachMoj:
Prove that we can read multiple "upstream" API specs, extract subsets of them, and republish as a combined "downstream" API spec.

Endpoint republishing is declared as follows...
```
    apiSpec.republish("activities.json", "/integration-api/", "/v2/activities/", tag="Activities")
    apiSpec.republish("plp.json", "/inductions/", "/v2/plp/inductions/", tag="Inductions")
    apiSpec.republish("plp.json", "/action-plans/", "/v2/plp/action-plans/", tag="Action Plans")
```
The parameters are...

1. The name of the cached upstream API spec
2. An upstream API path prefix (only paths with this prefix will be republished)
3. An external API path prefix (the upstream prefix will be replaced with this)
4. The tag (API doc section) to publish under

See https://dsdmoj.atlassian.net/browse/HIA-1139


pmcphee77:
PR is old. Closing. New PR will be created for this ticket HIA-1139

PR #1655, opened 2026-03-31
HIA-1120 POC for servlet filter test refactoring

SteveLeachMoj:
Introduce a new pattern for testing servlet filters. Our existing tests for filters do the following...

1. Setup the filter to be tested
2. Invoke that particular filter
3. Verify that the filter implementation called the expected methods on other objects

With these new tests we...

1. Set up the request, response and filter chain
2. Call the filter chain
3. Verify that the request and response match expected values

This allows us to refactor the filters in the filter chain, including moving functionality between filters, while preserving the behaviour of the chain as a whole and without breaking existing tests.

Importantly, these tests do _not_ assert that a specific filter called a specific function.


src/test/kotlin/uk/gov/justice/digital/hmpps/hmppsintegrationapi/extensions/AuthorisationFilterTest.kt
```diff
@@ -130,16 +199,163 @@ class AuthorisationFilterTest {
...
+    mockFilterChain(authorisationFilter).doFilter(mockRequest, resp)
+    assertThat(resp.errorMessage).isNotEqualTo("Certificate with serial number TEST_SERIAL_NUMBER has been revoked")
+  }
+
+  @Test
+  fun `ConsumerNameExtractionFilter calls the onward chain`() {
+    // From ConsumerNameExtractionFilterTest
+    val resp = MockHttpServletResponse()
+    val finalFilter = mock(Filter::class.java)
+    val req = mockRequest("GET", examplePath, "O=test,CN=sam", "9572494320151578633330348943480876283449388176")
```
jessicawyldemoj: This is a long string of a magic number, can we move this into a const so its clear what this number function servers

src/test/kotlin/uk/gov/justice/digital/hmpps/hmppsintegrationapi/extensions/AuthorisationFilterTest.kt
```diff
@@ -130,16 +199,163 @@ class AuthorisationFilterTest {
...
+    // From ConsumerNameExtractionFilterTest
+    val resp = MockHttpServletResponse()
+    val req = mockRequest("GET", examplePath, "O=test,CN=sam", "9572494320151578633330348943480876283449388176")
+
+    val chain = fullMockFilterChain(authorisationConfig)
+
+    chain.doFilter(req, resp)
+
+    assertThat(req.getAttribute("clientName")).isEqualTo("sam")
+    assertThat(req.getAttribute("certificateSerialNumber")).isEqualTo("01:AD:3E:D8:7D:D5:AA:84:F5:2D:83:E7:87:E9:90:E4:84:C5:2C:90")
```
jessicawyldemoj: Same with this long string too, we can move it into a const also

src/test/kotlin/uk/gov/justice/digital/hmpps/hmppsintegrationapi/extensions/AuthorisationFilterTest.kt
```diff
@@ -130,16 +199,163 @@ class AuthorisationFilterTest {
...
+
+    val req = mockRequest("GET", examplePath, "O=test,CN=consumer-name")
+
+    val chain = fullMockFilterChain(authConfig)
+
+    chain.doFilter(req, resp)
+
+    assertThat(req.getAttribute("filters")).isEqualTo(ConsumerFilters(prisons = listOf("MDI", "RolePrison")))
+
+    unmockkStatic("uk.gov.justice.digital.hmpps.hmppsintegrationapi.models.roleconfig.RoleKt")
```
jessicawyldemoj, line 344: What is going on here?

SteveLeachMoj: This is because the `AuthorisationConfig` (and other code) accesses the defined roles using a global static variable. See [PR 1656](https://github.com/ministryofjustice/hmpps-integration-api/pull/1656) for a failed attempt to fix this.

src/test/kotlin/uk/gov/justice/digital/hmpps/hmppsintegrationapi/extensions/AuthorisationFilterTest.kt
SteveLeachMoj: Done

src/test/kotlin/uk/gov/justice/digital/hmpps/hmppsintegrationapi/extensions/AuthorisationFilterTest.kt
SteveLeachMoj: Done

jessicawyldemoj (approved):

PR #1656, opened 2026-04-01
HIA-1120 Non-static access to defined roles by Auth Filter

SteveLeachMoj:
The `AuthorisationFilter` made direct use of the static `roles` list when matching consumers to roles which made it more difficult to test alternative role definitions. In order to minimise the need for static mocking, this PR switches to a non-static copy of the defined roles map.


SteveLeachMoj:
This is being done in https://github.com/ministryofjustice/hmpps-integration-api/pull/1757

PR #1657, opened 2026-04-01
HIA-1300: Add filters JSON object containing supervisionStatus

pmcphee77:
**THIS PR WILL JUST LOG THE OUTCOME OF THE SUPERVISION STATUS CHECK AND WILL NOT ACTUALLY USE THE SUPERVISION STATUS IN THE ELIGIBILITY CHECK. A FOLLOW UP PR WILL BE RAISED TO DO THIS** 

Introduced a new feature flag `include-supervision-status-attribute` on in `dev`, off in `preprod` and `prod`

When this feature flag is enabled:
1. The supervision status is determined using the hmppsId in a new `getSupervisionStatus` function in the `DomainEventIdentitiesResolver`. This makes use of the existing `getPersonSupervisionStatus` function in `GetPersonService` which has been made public
3. The supervisionStatus is be stored in a new `filters` field (json text) in the `event_notification` table
4. The supervisionStatus is checked against the consumers filters in the direct SQS send function to determine whether to send the message to the consumers queue. THIS PR WILL JUST LOG THE OUTCOME OF THE SUPERVISION STATUS CHECK AND WILL NOT ACTUALLY USE THE SUPERVISION STATUS IN THE ELIGIBILITY CHECK


src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsintegrationapi/events/entities/EventNotification.kt
```diff
@@ -14,6 +14,11 @@ data class EventNotification(
   val status: String? = "PENDING",
   @JsonProperty("lastModifiedDateTime")
   val lastModifiedDatetime: LocalDateTime? = null,
+  val filters: Filters? = null,
+)
+
+data class Filters(
```
SteveLeachMoj: Maybe `FilterValues` or `Metadata` rather than `Filters`, as this will contain the values that can be filtered on rather than the filtering to be done.

pmcphee77: Changed to `Metadata`

src/test/kotlin/uk/gov/justice/digital/hmpps/hmppsintegrationapi/extensions/MetadataExtractionFilterTest.kt
```diff
@@ -19,7 +19,7 @@ import uk.gov.justice.digital.hmpps.hmppsintegrationapi.models.roleconfig.Consum
 import uk.gov.justice.digital.hmpps.hmppsintegrationapi.models.roleconfig.Role
 import uk.gov.justice.digital.hmpps.hmppsintegrationapi.models.roleconfig.roles
 
-class FiltersExtractionFilterTest {
+class MetadataExtractionFilterTest {
```
SteveLeachMoj: I think this one actually is about filters rather than metadata isn't it?

src/main/resources/db/migration/V1_0_9__add_supervisionStatus_to_event_notification.sql
```diff
@@ -0,0 +1,2 @@
+ALTER TABLE EVENT_NOTIFICATION
+ADD METADATA varchar;
```
jessicawyldemoj, line 2: Question, will this new file be nullable/allow empty? Not too familiar with with this style of database migrations

src/test/kotlin/uk/gov/justice/digital/hmpps/hmppsintegrationapi/integration/events/listener/DomainEventsListenerIntegrationTest.kt
```diff
@@ -40,7 +40,7 @@ import java.time.Duration
 class DomainEventsListenerIntegrationTest : IntegrationTestWithEventsQueueBase() {
   val prisonId = "MDI"
 
-  val awaitTimeOut = Duration.ofSeconds(5)
+  val awaitTimeOut = Duration.ofSeconds(500)
```
jessicawyldemoj: What's with the timeout change?

src/test/kotlin/uk/gov/justice/digital/hmpps/hmppsintegrationapi/extensions/MetadataExtractionFilterTest.kt
pmcphee77: ahh yes

src/main/resources/db/migration/V1_0_9__add_supervisionStatus_to_event_notification.sql
pmcphee77: Yes its nullable and allows empty. If we didnt want it nullable we would explicitly set NOT NULL on it

src/test/kotlin/uk/gov/justice/digital/hmpps/hmppsintegrationapi/integration/events/listener/DomainEventsListenerIntegrationTest.kt
pmcphee77: Good point and good spot - reverted that

src/main/resources/db/migration/V1_0_9__add_supervisionStatus_to_event_notification.sql
jessicawyldemoj: Good to note for the future!

jessicawyldemoj (approved):

PR #1676, opened 2026-04-09
HIA-1260: Added SQS wrapper for testing

pmcphee77:
Created a `QueueService` to enable different implementations of the queue processing:
`AwsQueueService` uses SQS via the `HMPPSQueueService`
`TestQueueService` saves messages in memory


src/test/kotlin/uk/gov/justice/digital/hmpps/hmppsintegrationapi/util/TestQueue.kt
```diff
@@ -0,0 +1,42 @@
...
+class TestQueue(
+  val queueName: String,
+) {
+  var messages = mutableListOf<String>()
+
+  fun publish(message: String) {
+    messages.add(message)
+  }
+
+  inline fun <reified T : Any> messagesOnQueue(): List<T> =
```
SteveLeachMoj: `publish` and `messagesOnQueue` look like they are asymmetric, in that `publish` takes a (pre-marshalled) string and `messagesOnQueue` returns a (pre-unmarshalled) list of objects. It might be better if both worked on strings, or both worked on objects.

pmcphee77: This has been changed now.
`messages` returns a list of strings
`publish` takes a string
`messagesAsObjects<T>()` can be used to marshal into objects

jessicawyldemoj (approved):

PR #1677, opened 2026-04-13
HIA-1329: Updated gradle

jessicawyldemoj:


pmcphee77 (commented):
we need to upgrade the GHA code changing java-version: 21 to java-version: 25

.github/workflows/security-scanning.yml
```diff
@@ -22,7 +22,7 @@ jobs:
         uses: actions/setup-java@be666c2fcd27ec809703dec50e508c2fdc7f6654 # v5
         with:
           distribution: 'temurin'
-          java-version: '21'
+          java-version: '25'
```
pmcphee77: Should we remove the quotes, so we can sopt these easier in the future?

pmcphee77 (approved):

PR #1689, opened 2026-04-20
HIA-1300: Added a supervision status by hmppsId function

pmcphee77:
We are currently running the events supervision checks in "logging only" in DEV only so we can check that the events are resolving to the correct supervision status before we turn it on in dev.

We found that the supervision status is NOT being determined correctly.
Therefore Steve created the following pseudo code here: https://dsdmoj.atlassian.net/wiki/spaces/HIA/pages/5983084466/Supervision+Status#Supervision-Status-derivation-(tactical)

This PR introduces a new function in GetPersonService (getSupervisionStatus(hmppsId) which will get the supervision status for a given hmppsId

This is now as follows:
```
FETCH Offender FROM NDeliusGateway
IF we have a NOMIS number
  FETCH POSPrisoner FROM PrisonerOffenderSearchGateway
IF POSPrisoner exists AND POSPrisoner.status starts with "ACTIVE" 
  SupervisionStatus = "PRISONS"
ELSE IF Offender.activeProbationManagedSentence IS TRUE
  SupervisionStatus = "PROBATION"
ELSE 
  SupervisionStatus = "NONE"
If we cannot access the required fields for the offender but need to proceed anyway then we set SupervisionStatus to “UNKNOWN”.
```

### DIFF
--- a/src/test/kotlin/uk/gov/justice/digital/hmpps/hmppsintegrationapi/services/GetPersonServiceTest.kt
+++ b/src/test/kotlin/uk/gov/justice/digital/hmpps/hmppsintegrationapi/services/GetPersonServiceTest.kt
@@ -1079,7 +1079,7 @@ internal class GetPersonServiceTest(
           result.shouldBe(SupervisionStatus.PROBATION)
         }
 
-        it("HmppsId resolves to a supervision status of PROBATION") {
+        it("HmppsId resolves to a supervision status of NONE") {
           whenever(deliusGateway.getOffender(any())).thenReturn(Response(data = Offender("Test", "Test", activeProbationManagedSentence = false, otherIds = OtherIds(crn = crnNumber, nomsNumber = nomsNumber))))
           val result = getPersonService.getSupervisionStatus(crnNumber)
           result.shouldBe(SupervisionStatus.NONE)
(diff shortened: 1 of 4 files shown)


src/test/kotlin/uk/gov/justice/digital/hmpps/hmppsintegrationapi/services/GetPersonServiceTest.kt
```diff
@@ -1071,5 +1071,61 @@ internal class GetPersonServiceTest(
...
       }
+
+      describe("get Supervision Status by hmppsId") {
+        it("HmppsId resolves to a supervision status of PROBATION") {
+          whenever(deliusGateway.getOffender(any())).thenReturn(Response(data = Offender("Test", "Test", activeProbationManagedSentence = true, otherIds = OtherIds(crn = crnNumber, nomsNumber = nomsNumber))))
+          val result = getPersonService.getSupervisionStatus(crnNumber)
+          result.shouldBe(SupervisionStatus.PROBATION)
+        }
+
+        it("HmppsId resolves to a supervision status of PROBATION") {
```
jessicawyldemoj: Test says outcome should be PROBATION but is expecting NONE

pmcphee77: Great spot - changed

jessicawyldemoj (approved):

PR #1705, opened 2026-04-24
HIA-1345: Given feature flags to default to true as well as false.

jessicawyldemoj:


src/test/kotlin/uk/gov/justice/digital/hmpps/hmppsintegrationapi/services/GetPersonServiceTest.kt
```diff
@@ -60,18 +59,14 @@ import uk.gov.justice.digital.hmpps.hmppsintegrationapi.roles.dsl.SupervisionSta
...
-  @MockitoBean val deliusGateway: NDeliusGateway,
-  @MockitoBean val corePersonRecordGateway: CorePersonRecordGateway,
-  @MockitoBean val featureFlagConfig: FeatureFlagConfig,
-  @MockitoBean val telemetryService: TelemetryService,
-  private val getPersonService: GetPersonService,
+  val prisonerOffenderSearchGateway: PrisonerOffenderSearchGateway = mock(),
+  val consumerPrisonAccessService: ConsumerPrisonAccessService = mock(),
+  val deliusGateway: NDeliusGateway = mock(),
+  val corePersonRecordGateway: CorePersonRecordGateway = mock(),
+  val featureFlagConfig: FeatureFlagConfig = spy<FeatureFlagConfig>(),
```
SteveLeachMoj: Do we really need a spy here? Can't we just set the flags to what we want them to be for the test?

jessicawyldemoj: This is here as there is one flag test later down to actually see if the flag works

SteveLeachMoj (approved):

PR #1708, opened 2026-04-24
HIA-1352: Added logging to detect x-on-behalf-of from the headers

jessicawyldemoj:


scripts/K6/load-tests.js
```diff
@@ -67,6 +67,7 @@ function call_status_endpoint(path) {
     headers: {
       'Content-Type': 'application/json',
       'x-api-key': api_key,
+      'X-On-Behalf-Of': 'smoketest'
```
SteveLeachMoj: Don't think we need this

scripts/K6/smoke-tests.js
```diff
@@ -40,6 +40,7 @@ const httpParams = {
   headers: {
     'Content-Type': 'application/json',
     'x-api-key': api_key,
+    'X-On-Behalf-Of': 'smoketest'
```
SteveLeachMoj: I think this will add the header to all smoke test calls, but it might be better to just add a new call that is a copy of an existing one, but with the additional header.

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsintegrationapi/extensions/AuthorisationFilter.kt
```diff
@@ -41,6 +44,11 @@ class AuthorisationFilter(
       return
     }
 
+    val onBehalfOf = req.getAttribute("onBehalfOf") as String?
+    if (onBehalfOf != null) {
+      log.info("Header found with X-On-Behalf-Of with the value '$onBehalfOf'")
```
SteveLeachMoj: I don't think this is needed if we're including it in the span attributes

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsintegrationapi/extensions/ConsumerNameExtractionFilter.kt
```diff
@@ -26,9 +26,11 @@ class ConsumerNameExtractionFilter : Filter {
     response as HttpServletResponse
     val subjectDistinguishedName = req.getHeader("subject-distinguished-name")
     val certificateSerialNumber = extractCertificateSerialNumber(req.getHeader("cert-serial-number"))
+    val onBehalfOf = req.getHeader("X-On-Behalf-Of")
     val extractedConsumerName = extractConsumerName(subjectDistinguishedName)
     req.setAttribute("clientName", extractedConsumerName)
     req.setAttribute("certificateSerialNumber", certificateSerialNumber)
+    req.setAttribute("onBehalfOf", onBehalfOf)
```
SteveLeachMoj: I'm not sure we need to do this, at least for this ticket, and we should probably be avoiding changes to this filter given that we want to get rid of it.

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsintegrationapi/config/telemetry/ClientTrackingInterceptor.kt
```diff
@@ -40,6 +40,12 @@ class ClientTrackingInterceptor : HandlerInterceptor {
       runCatching { Span.current().setAttribute("certSerialNumber", it) }
     }
 
+    // Set X-On-Behalf-Of in app insights
+    val onBehalfOf = request.getAttribute("onBehalfOf") as String?
```
SteveLeachMoj: Can't we just fetch the value directly from the header?

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsintegrationapi/config/telemetry/ClientTrackingInterceptor.kt
```diff
@@ -19,27 +19,27 @@ class ClientTrackingConfiguration(
...
-        // Do nothing - don't create client id span
-      }
+      telemetryService.setSpanAttribute("clientId", it)
     }
-
     // Set the certificate serial number in app insights
     val certificateSerialNumber = request.getAttribute("certificateSerialNumber") as String?
     certificateSerialNumber?.let {
-      runCatching { Span.current().setAttribute("certSerialNumber", it) }
+      runCatching { telemetryService.setSpanAttribute("certSerialNumber", it) }
```
SteveLeachMoj: Don't need the runCatching as that's now handled by setSpanAttribute

scripts/K6/smoke-tests.js
```diff
@@ -654,6 +654,26 @@ function verify_prisoner_contacts(hmppsId) {
...
+function verify_headers() {
+  //Tests to add headers too for app insights
+  httpParams = {
+    headers: {
+      'Content-Type': 'application/json',
+      'x-api-key': api_key,
+      'X-On-Behalf-Of': 'Test-Behalf-Of'
+    },
+  };
+  let res = validate_get_request(`/v1/ping`)
```
SteveLeachMoj: Is this an actual endpoint? Maybe `/v1/status` is a better option?

scripts/K6/smoke-tests.js
```diff
@@ -654,6 +654,26 @@ function verify_prisoner_contacts(hmppsId) {
...
+
+  // Reset headers for future tests
+  httpParams = {
```
SteveLeachMoj: Rather than setting and restoring the `httpParams` variable, maybe just call `http.get()` directly rather than via the `validate_get_request()` wrapper function.

src/test/kotlin/uk/gov/justice/digital/hmpps/hmppsintegrationapi/config/telemetry/ClientTrackingInterceptorTest.kt
```diff
@@ -6,11 +6,13 @@ import org.junit.jupiter.api.Assertions.assertTrue
 import org.junit.jupiter.api.Test
 import org.mockito.Mockito.mock
 import org.mockito.kotlin.whenever
+import uk.gov.justice.digital.hmpps.hmppsintegrationapi.telemetry.TelemetryService
 
 class ClientTrackingInterceptorTest {
   private val request: HttpServletRequest = mock(HttpServletRequest::class.java)
```
SteveLeachMoj, line 14: Not really one for this PR because we're just repeating an existing pattern, but Spring already provides "mock" versions of HttpServletRequest and HttpServletResponse.

https://docs.spring.io/spring-framework/docs/current/javadoc-api/org/springframework/mock/web/MockHttpServletRequest.html

scripts/K6/smoke-tests.js
jessicawyldemoj: actually double checking this actual endpoint is /v1/health/ping (so this is wrong) . Any endpoint here is good to use though

src/test/kotlin/uk/gov/justice/digital/hmpps/hmppsintegrationapi/config/telemetry/ClientTrackingInterceptorTest.kt
jessicawyldemoj: I agree on this, created a ticket here to do a shift like that in one go https://dsdmoj.atlassian.net/jira/software/c/projects/HIA/boards/5033?quickFilter=17109&selectedIssue=HIA-1377

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsintegrationapi/config/telemetry/ClientTrackingInterceptor.kt
```diff
@@ -19,27 +19,28 @@ class ClientTrackingConfiguration(
...
     }
-
     // Set the certificate serial number in app insights
     val certificateSerialNumber = request.getAttribute("certificateSerialNumber") as String?
     certificateSerialNumber?.let {
-      runCatching { Span.current().setAttribute("certSerialNumber", it) }
+      telemetryService.setSpanAttribute("certSerialNumber", it)
+    }
+    // Set on behalf off in app insights
+    val onBehalfOff = request.getHeader("X-On-Behalf-Of")
```
SteveLeachMoj: Very minor typo: "Of" not "Off"

SteveLeachMoj (approved):

PR #1749, opened 2026-05-12
HIA-1120: Consolidate filters and make roles more testable

pmcphee77:
Moved processing to `AuthorisationFilter` and deleted the following:
`ClientTrackingInterceptor`
`ConsumerNameExtractionFilter`
`FiltersExtractionFilter`

In order to improve testability.
Changed `AuthorisationConfig` to be a more of a service that is provided with roles and an authorisation config and moved the `ConfigurationProperties` binding with the authorisation config to a standalone class `Config`. 
Renaming at this stage would cause confusion

This has enabled me to move all of the tests from the `FiltersExtractionFilterTest` that relied on `mockkstatic` to the `AuthorisationFilterTest` without the need of `mockkstatic`


src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsintegrationapi/controllers/v2/ConfigController.kt
SteveLeachMoj, line 1: Just a note that I don't think we need this controller any more, it was just used to give `hmpps-integration-events` access to the config from `hmpps-integration-api`.

jessicawyldemoj (approved):

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsintegrationapi/controllers/v2/ConfigController.kt
pmcphee77: Ive raised another ticket for this HIA-1419

PR #1751, opened 2026-05-13
Hia 1120 consolidate filters 2

pmcphee77:
PR that reintroduces the revert of this ticket along with the local implementation of the AuthorisationFilter

Moved processing to `AuthorisationFilter` and deleted the following:
`ClientTrackingInterceptor`
`ConsumerNameExtractionFilter`
`FiltersExtractionFilter`

Created an `AuthorisationService` and moved functions from `AuthorisationConfig`


src/main/resources/application-local-docker.yml
```diff
@@ -51,6 +51,7 @@ authorisation:
       roles:
         - "full-access"
       filters:
+  default-consumer-name: "all-access"
```
jessicawyldemoj, line 54: you sure this shouldn't be "full-access" instead?

pmcphee77: yes - im sure - thats the consumer name, not the role name

jessicawyldemoj (approved):

PR #1754, opened 2026-05-13
HIA-1372: Added jwt logging

jessicawyldemoj:
Code to start HIA-1372.

This ticket only logs different values found in the jwt and logs them.

As well this logs the `unique_name` as part of the request.


pmcphee77:
The reason the Controller tests arent working is that these use the `WebMvcTestConfiguration` which loads the beans that `WebMvcTest` doesnt.

So in that class just add this
```
  @Bean
  @ConditionalOnMissingBean
  fun oboService(): OboService = OboService()
```

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsintegrationapi/services/OboService.kt
```diff
@@ -0,0 +1,22 @@
...
+
+@Component
+class OboService {
+  fun decodeJwt(jwtToken: String?): Claims? =
+    try {
+      val jwtDecoded =
+        Jwts
+          .parser()
+          .build()
+          .parse(jwtToken) as Jws<Claims>
```
pmcphee77: Does this parse(jwtToken) actually return a Claims? I think if we want to return a claim then we should use the parseClaimsJwt (which is deprecated) so parseUnsecuredClaims instead?

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsintegrationapi/services/OboService.kt
```diff
@@ -0,0 +1,22 @@
...
+      jwtDecoded.payload["kid"] = jwtDecoded.header["kid"] // Add kid value as thats in the header not playload
```
pmcphee77: payload not playload

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsintegrationapi/services/OboService.kt
jessicawyldemoj: So I can confirm it returns claims, and I am happy to switch over to that parser instead.

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsintegrationapi/services/onBehalfOf/EntraJwtOboService.kt
```diff
@@ -0,0 +1,21 @@
...
+@Component
+class OboService {
+  fun decodeJwt(jwtToken: String?): Claims? =
+    try {
+      val jwtDecoded =
+        Jwts
+          .parser()
+          .build()
+          .parseUnsecuredClaims(jwtToken)
+      jwtDecoded.payload["kid"] = jwtDecoded.header["kid"] // Add kid value as that's in the header not payload
```
pmcphee77: Just wondering why we are adding this into the payload?

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsintegrationapi/models/roleconfig/ConsumerConfig.kt
```diff
@@ -6,6 +6,7 @@ data class ConsumerConfig(
   val roles: List<String>? = null,
   val notes: String? = null,
   val queueName: String? = null,
+  val oboConfig: String? = null,
```
pmcphee77: Are we not using this yet? Should this be set on the smoke-test consumer?

SteveLeachMoj: Shouldn't it also be on the `jaiu-onboarding-optimisation` consumer in dev?

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsintegrationapi/services/OboService.kt
```diff
@@ -0,0 +1,21 @@
+package uk.gov.justice.digital.hmpps.hmppsintegrationapi.services
+
+import io.jsonwebtoken.Claims
+import io.jsonwebtoken.Jwts
+import org.springframework.stereotype.Component
+
+@Component
+class OboService {
+  fun decodeJwt(jwtToken: String?): Claims? =
```
pmcphee77: Could we make `jwtToken` a non nullable field? And only call it from the filter if the jwt is in the header?

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsintegrationapi/extensions/AuthorisationFilter.kt
```diff
@@ -55,8 +57,23 @@ class AuthorisationFilter(
     // Get the on behalf of token
     val onBehalfOf = req.getHeader("X-On-Behalf-Of")
 
+    val decodedJwt = oboService.decodeJwt(onBehalfOf)
+
+    // capture token if decoded
+    if (decodedJwt != null) {
+      telemetryService.captureMessage(
```
SteveLeachMoj: Is there a particular reason for using the telemetry service for this, rather than just logging it?

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsintegrationapi/extensions/AuthorisationFilter.kt
```diff
@@ -55,8 +57,23 @@ class AuthorisationFilter(
...
+        "Token found with iss:${decodedJwt["iss"]}, " +
+          "appId:${decodedJwt["appid"]}, " +
+          "unique_name:${decodedJwt["unique_name"]}, " +
+          "kid:${decodedJwt["kid"]}, " +
+          "nbf:${decodedJwt["nbf"]}, " +
+          "exp:${decodedJwt["exp"]}, " +
+          "aud:${decodedJwt["aud"]}",
+      )
```
pmcphee77: Is there any reasons why we dont just log in the oboService itself? and just inject the telemetry service to that

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsintegrationapi/models/roleconfig/ConsumerConfig.kt
jessicawyldemoj: Ah this is not for this pr, I have work ready for other tickets later ready just as I remember to do stuff, this is going to be removed and added in a later pr

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsintegrationapi/services/onBehalfOf/EntraJwtOboService.kt
jessicawyldemoj: The header and the payload are two different lists in the decodedJwt object, we just want to combine the list into a list of claims we can later use

scripts/K6/smoke-tests.js
```diff
@@ -661,7 +661,27 @@ function verify_headers() {
...
+//          "iat":1516239022
+//          "aud":"testAud"
+//          "iss":"testIss"
+//          "nbf":1778083253
+//          "exp":1978087263
+//          "appid":"testId"
+//          "unique_name":"testName"
+//        }
+//      }
+      'X-On-Behalf-Of': 'eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCIsImtpZCI6InRlc3RLaWQifQ.eyJzdWIiOiIxMjM0NTY3ODkwIiwibmFtZSI6IkpvaG4gRG9lIiwiYWRtaW4iOnRydWUsImlhdCI6MTUxNjIzOTAyMiwiYXVkIjoidGVzdEF1ZCIsImlzcyI6InRlc3RJc3MiLCJuYmYiOjE3NzgwODMyNTMsImV4cCI6MTk3ODA4NzI2MywiYXBwaWQiOiJ0ZXN0SWQiLCJ1bmlxdWVfbmFtZSI6InRlc3ROYW1lIn0.bppq2M56ruzzxwfEWK408np-22hAJ2vwrHlbHGuWDq0'
```
SteveLeachMoj: I suspect that the fixed dates in this will mean it fails to validate

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsintegrationapi/services/OboService.kt
```diff
@@ -0,0 +1,21 @@
+package uk.gov.justice.digital.hmpps.hmppsintegrationapi.services
+
+import io.jsonwebtoken.Claims
+import io.jsonwebtoken.Jwts
+import org.springframework.stereotype.Component
+
+@Component
+class OboService {
```
SteveLeachMoj: I think we wanted `OboService` to be an interface, with `EntraJwtOboService` (or something) as a concrete implementation of that interface.

jessicawyldemoj: I remember agreeing to the opposite, but if that was what an actually agreed upon then that is an easy fix. Tbh its not important here as this ticket is just to show decoding the token and logging it but a simple change we can do now

scripts/K6/smoke-tests.js
jessicawyldemoj: It will yes, but just for logging proposes we do just want to add it now to see if it actually saves the date in app insights. We should remove the expiry date later and have a bare jwt for smoke tests

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsintegrationapi/extensions/AuthorisationFilter.kt
jessicawyldemoj: We can do that yes

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsintegrationapi/services/OboService.kt
jessicawyldemoj: Sure thing

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsintegrationapi/services/OboService.kt
SteveLeachMoj: I thought the point of the `oboConfig` value was to specify which `OboService` implementation should be used?

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsintegrationapi/extensions/AuthorisationFilter.kt
SteveLeachMoj: Why do we need the telemetry service for this? We can just log it. 

The telemetry service is useful for sending to Sentry, or for enriching the span information in AppInsights, but isn't needed for basic logging is it?

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsintegrationapi/services/onBehalfOf/EntraJwtOboService.kt
```diff
@@ -0,0 +1,19 @@
...
+  override fun decodeJwt(jwtToken: String): Claims? =
+    try {
+      val jwtDecoded =
+        Jwts
+          .parser()
+          .build()
+          .parseUnsecuredClaims(jwtToken)
+      jwtDecoded.payload["kid"] = jwtDecoded.header["kid"] // Add kid value as that's in the header not payload
+      jwtDecoded.payload
+    } catch (e: Exception) {
```
SteveLeachMoj: This will silently swallow all the information about the error won't it?

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsintegrationapi/extensions/AuthorisationFilter.kt
```diff
@@ -189,11 +195,13 @@ class AuthorisationFilter(
     clientId: String,
     certSerialNumber: String?,
     onBehalfOf: String?,
+    uniqueName: String?,
```
SteveLeachMoj: I'm not sure we need to include both of these. If we can extract the uniqueName then we probably don't need to log the header value.

pmcphee77 (approved):

PR #1759, opened 2026-05-18
HIA-1372: Added obo config as well as improved unit tests

jessicawyldemoj:


src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsintegrationapi/models/roleconfig/ConsumerConfig.kt
```diff
@@ -6,6 +6,7 @@ data class ConsumerConfig(
   val roles: List<String>? = null,
   val notes: String? = null,
   val queueName: String? = null,
+  val oboConfig: String? = null,
```
SteveLeachMoj: I don't know if it is worth doing it in this PR or leaving it until a later one, but we're probably going to need more configuration than just which OBO service to use. Maybe make this an `OboConfig` object with a `strategy` field or something? Adding a new field to an existing data structure is a backwards-compatible change, but converting a string to a data structure isn't.

src/main/resources/application-test.yml
```diff
@@ -171,6 +171,12 @@ authorisation:
         supervisionStatuses:
           - PROBATION
       queueName: "testqueue"
+    obo-unsigned:
+      roles:
+        - "all-endpoints"
+      oboConfig:
+        strategy:
+          - "unsigned"
```
SteveLeachMoj: `strategy` can be a string rather than a list of strings

pmcphee77 (dismissed):

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsintegrationapi/models/oboconfig/OboConfig.kt
```diff
@@ -0,0 +1,5 @@
+package uk.gov.justice.digital.hmpps.hmppsintegrationapi.models.oboconfig
+
+data class OboConfig(
+  val strategy: String? = null,
```
SteveLeachMoj: I can't think of a reason why `strategy` would ever be null.

src/test/kotlin/uk/gov/justice/digital/hmpps/hmppsintegrationapi/integration/OnBehalfOfIntegrationTest.kt
```diff
@@ -0,0 +1,65 @@
...
+
+  @Test
+  fun `if oboConfig is empty, do not call any obo service`() {
+    whenever(mockRequest.getHeader("subject-distinguished-name")).thenReturn("C=GB,ST=London,L=London,O=Home Office,CN=obo-empty")
+    filtersExtractionFilter.doFilter(mockRequest, mockResponse, mockChain)
+    verify(authorisationService, times(1)).oboService("obo-empty")
+  }
+
+  @Test
+  fun `if oboConfig is unassigned, call the unassigned obo service`() {
```
SteveLeachMoj: "unsigned" rather than "unassigned"

src/test/kotlin/uk/gov/justice/digital/hmpps/hmppsintegrationapi/integration/OnBehalfOfIntegrationTest.kt
```diff
@@ -0,0 +1,65 @@
...
+import org.mockito.Mockito.verify
+import org.mockito.kotlin.whenever
+import org.springframework.boot.webmvc.test.autoconfigure.AutoConfigureMockMvc
+import uk.gov.justice.digital.hmpps.hmppsintegrationapi.extensions.AuthorisationFilter
+import uk.gov.justice.digital.hmpps.hmppsintegrationapi.services.onbehalfof.createUnassignedJwy
+import uk.gov.justice.digital.hmpps.hmppsintegrationapi.telemetry.TelemetryService
+import kotlin.test.Test
+
+@AutoConfigureMockMvc
+class OnBehalfOfIntegrationTest : IntegrationTestBase() {
```
SteveLeachMoj, line 7: Why is this an integration test, out of interest? It doesn't use the database or any upstream APIs from what I can see.

jessicawyldemoj: As far as I am aware integration tests are the way to test if config load is working by file. So it's to test if the new config actually works, if I am wrong do let me know.

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsintegrationapi/models/oboconfig/OboConfig.kt
jessicawyldemoj: Common pattern in config we have. But you are right if OboConfig is created it should have a strategy so we can make it non nullable

src/test/kotlin/uk/gov/justice/digital/hmpps/hmppsintegrationapi/integration/OnBehalfOfIntegrationTest.kt
```diff
@@ -0,0 +1,65 @@
...
+  val mockRequest = mock(HttpServletRequest::class.java)
+  val mockResponse = mock(HttpServletResponse::class.java)
+  val mockChain = mock(FilterChain::class.java)
+  val mockTelemetryService = mock(TelemetryService::class.java)
+
+  lateinit var filtersExtractionFilter: AuthorisationFilter
```
pmcphee77: I think this should be renamed from `filtersExtractionFilter` to `authorisationFilter`. I think this is a hangover from a refactor

src/test/kotlin/uk/gov/justice/digital/hmpps/hmppsintegrationapi/integration/OnBehalfOfIntegrationTest.kt
```diff
@@ -0,0 +1,55 @@
...
+  }
+
+  @Test
+  fun `if oboConfig is invalid, and no Jwt is provided, return unauthorized`() {
+    callApiWithCN("/v1/status", "obo-invalid")
+      .andExpect(MockMvcResultMatchers.status().isUnauthorized)
+  }
+
+  @Test
+  fun `if oboConfig is invalid, and a Jwt is provided, return ok`() {
```
SteveLeachMoj: this returns unauthorised - test is OK, description is wrong

jessicawyldemoj: Ah good catch, yes the test description should say unauthorised

pmcphee77 (approved):

PR #1768, opened 2026-05-21
HIA-1372 Validate JWT using JWKS

SteveLeachMoj:
Introduce a `JwksOboService` implementation that loads the public keys for signature validation from a JWKS file, and use it for JWTs signed by the MoJ Entra. The Entra-specific configuration is currently hard-coded in the `AuthorisationService` but could be derived dynamically if needed in future.

The unit tests for `JwksOboService` use a local JWKS and so are entirely self-contained.

https://dsdmoj.atlassian.net/browse/HIA-1406

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsintegrationapi/services/AuthorisationService.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsintegrationapi/services/AuthorisationService.kt
@@ -181,12 +181,13 @@ class AuthorisationService(
   fun allNull(vararg values: List<Any>?) = values.all { it == null }
 
   private val entraJwksUrl = URI.create("https://login.microsoftonline.com/common/discovery/v2.0/keys").toURL()
+  private val entraOboService = JwksOboService(entraJwksUrl, "unique_name")
 
   fun oboService(consumerName: String): OboService? {
     val oboServiceName = authorisationConfig.consumers[consumerName]?.oboConfig?.strategy
     return when (oboServiceName) {
       "unsigned" -> UnsignedJwtOboService()
-      "entra" -> JwksOboService(entraJwksUrl, "unique_name")
+      "entra" -> entraOboService // reuse the same instance to benefit from JWKS caching
       else -> null
     }
   }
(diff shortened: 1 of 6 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsintegrationapi/services/AuthorisationService.kt
```diff
@@ -178,11 +180,14 @@ class AuthorisationService(
...
 
+  private val entraJwksUrl = URI.create("https://login.microsoftonline.com/common/discovery/v2.0/keys").toURL()
+  private val entraOboService = JwksOboService(entraJwksUrl, "unique_name")
+
   fun oboService(consumerName: String): OboService? {
     val oboServiceName = authorisationConfig.consumers[consumerName]?.oboConfig?.strategy
     return when (oboServiceName) {
       "unsigned" -> UnsignedJwtOboService()
-      "entra" -> null // EntraJwtOboService()
+      "entra" -> entraOboService // reuse the same instance to benefit from JWKS caching
```
pmcphee77, line 190: How about
```
 "entra" -> {
        JwksOboService(
          jwks = URI.create("https://login.microsoftonline.com/common/discovery/v2.0/keys").toURL(), 
          usernameClaim = "unique_name"
        )
      }
```

And then we can remove
```
private val entraJwksUrl = URI.create("https://login.microsoftonline.com/common/discovery/v2.0/keys").toURL()
  private val entraOboService = JwksOboService(entraJwksUrl, "unique_name")
```

SteveLeachMoj: Will do

SteveLeachMoj: Actually we need a persistent instance for caching keys

pmcphee77: agreed

pmcphee77 (approved):

PR #1777, opened 2026-05-27
HIA-1353: Support for username in HMPPS Auth token requests

jessicawyldemoj:


src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsintegrationapi/gateways/HmppsAuthGateway.kt
```diff
@@ -80,8 +84,7 @@ class HmppsAuthGateway(
 
     telemetryService.trackEvent("AuthTokenRequest")
     val credentials = Credentials(username, password)
-    val uri = "/auth/oauth/token?grant_type=client_credentials"
-
+    val uri = "/auth/oauth/token?grant_type=client_credentials${requestContext?.oboUserName?.let {"&username=${requestContext.oboUserName}"} ?: ""}"
```
SteveLeachMoj, line 61: I'm not personally a huge fan of complex expressions in string templates, but @pmcphee77  might be fine with it.

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsintegrationapi/services/GetPersonService.kt
```diff
@@ -398,13 +399,13 @@ class GetPersonService(
     val probationData = probationResponse.data
 
     // If supervisionStatus filter is Probation Only and probation record is NOT under active supervision
-    if (filters?.isProbationOnly() == true && probationData?.underActiveSupervision == false) {
+    if (requestContext?.filters?.isProbationOnly() == true && probationData?.underActiveSupervision == false) {
```
SteveLeachMoj, line 402: As mentioned on the call, I might be temped to add `val filters = requestContext?.filters` above this so that the rest of the code is unchanged.

PR #1788, opened 2026-06-01
HIA-1398: Removed the test consumer from the dev profile

pmcphee77:


src/main/resources/application-dev.yml
```diff
@@ -152,11 +152,6 @@ authorisation:
     smoke-test-no-access:
       notes: "Automated post-deployment smoke tests (all access denied)"
       roles:
-    test:
-      notes: "Used for testing queues in dev"
-      roles:
-        - "mappa"
-      queueName: "testclientsubscriberqueue"
```
jessicawyldemoj, line 159: By the looks of things this queue here is only used for this area. Do we also want to remove this queneName from the code too?

pmcphee77:
Closing as not clear what is required

PR #1798, opened 2026-06-03
HIA-1371: Fixed formatting to now be a bullet point list

jessicawyldemoj:


pmcphee77:
This PR still doesnt have the breaks

image

So we need the <br> on every line.
Also its best to use self closing tags so use
`</br>`

pmcphee77 (approved):

PR #1828, opened 2026-06-11
HIA-1411: Cert generation script improvements

jessicawyldemoj:
Update the cert generation script generate.sh to include a confirmation of input details before generation. 
Validate the CSR is the same as the consumer name.
Add user check to see if we want to upload to S3 bucket

Paul Testing
Testing invalid CSR
image

Testing valid CSR

image

Testing confirm with no upload

image

### DIFF
--- a/scripts/client_certificates/generate.sh
+++ b/scripts/client_certificates/generate.sh
@@ -179,7 +179,7 @@ read_certificate_arguments() {
 
   echo "Upload Backups to S3?:  $( [[ "$upload_to_s3" == 1 ]] && echo 'Yes' || echo 'No' )"
   echo "-----------------------------------------------------"
-  echo "Are you sure these options are correct? (y/n)"
+  echo "Do you wish to proceed? (y/n)"
   read confirmation
 
   case $confirmation in
@@ -209,8 +209,6 @@ generate_client() {
   else
     cp ./csrs/$environment/$client/$environment-$client-client.csr .
   fi
-
-  validate_csr "$environment-$client-client.csr"
 
   openssl x509 -req -in $environment-$client-client.csr -CA truststore.pem -CAkey truststore.key -out $environment-$client-client.pem -days 365 -sha256 -CAcreateserial
 }
--- a/scripts/client_certificates/generate.sh
+++ b/scripts/client_certificates/generate.sh
@@ -14,7 +14,7 @@ validate_csr() {
   fi
 
   local csr_cn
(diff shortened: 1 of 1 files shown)


scripts/client_certificates/generate.sh
```diff
@@ -5,6 +5,33 @@ clean() {
...
+validate_csr() {
+  local csr_file=$1
+
+  if [[ ! -f "$csr_file" ]]; then
+    echo "Error: CSR file $csr_file not found for validation."
+    exit 1
+  fi
+
+  local csr_cn
+  csr_cn=$(openssl req -in "$csr_file" -noout -subject | sed -n 's/.*[=,/]CN *=[ ]* \?\([^/,]*\).*/\1/p' | xargs)
```
jessicawyldemoj: If anyone has a way to make this line better I am all up for it, not so happy with the regex

scripts/client_certificates/generate.sh
```diff
@@ -112,6 +140,57 @@ read_certificate_arguments() {
...
+      echo "Add API Key to Tarball: $( [[ "$add_api_key" == 1 ]] && echo 'Yes' || echo 'No' )"
+    fi
+  else
+    echo "Client Organisation:    $organisation"
+    echo "Client Region:          $region"
+  fi
+
+  echo "Upload Backups to S3?:  $( [[ "$upload_to_s3" == 1 ]] && echo 'Yes' || echo 'No' )"
+  echo "-----------------------------------------------------"
+  echo "Are you sure these options are correct? (y/n)"
```
SteveLeachMoj: I'd make this "Do you wish to proceed?" so that it is clear that answering "yes" will actually go on and create the cert

scripts/client_certificates/generate.sh
```diff
@@ -130,6 +209,9 @@ generate_client() {
   else
     cp ./csrs/$environment/$client/$environment-$client-client.csr .
   fi
+
+  validate_csr "$environment-$client-client.csr"
```
pmcphee77: Im not sure we need the second validate_csr call in the generate_client  function TBH as this has already been done. We dont need to validate it again do we? and no point validating an automatically generated one?

scripts/client_certificates/generate.sh
pmcphee77: This didnt work for me.

`csr_cn=$(openssl req -in "$csr_file" -noout -subject | sed -n '/^subject/s/^.*CN=//p' | xargs)`

pmcphee77 (approved):

PR #1830, opened 2026-06-15
HIA-1459: Assessment summary change notifications

jessicawyldemoj:


src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsintegrationapi/events/enums/IntegrationEventType.kt
```diff
@@ -381,6 +381,11 @@ enum class IntegrationEventType(
     "v1/persons/{hmppsId}/risks/serious-harm",
     { ROSH_TYPES.contains(it.eventType) },
   ),
+  ASSESSMENT_SUMMARY_CHANGE(
+    "v1/persons/{hmppsId}/assessment-summary",
+    { ROSH_TYPES.contains(it.eventType) },
```
SteveLeachMoj: Shouldn't this be something like `ASSESSMENT_SUMMARY_TYPES` which happens to (currently) have the same events as `ROSH_TYPES`?

pmcphee77 (approved):

PR #1833, opened 2026-06-15
HIA-1493: contact search

pmcphee77:
PR to enable searching by contacts.


jessicawyldemoj:
Do we not want to add this new endpoint into Role.kt? As well for dev testing do you want to set your profile to the all endpoints role so we can gather dev evidence

jessicawyldemoj:
Don't we need this behind a feature flag?

pmcphee77:
> Don't we need this behind a feature flag?

Yes - done

pmcphee77:
> Do we not want to add this new endpoint into Role.kt? As well for dev testing do you want to set your profile to the all endpoints role so we can gather dev evidence

Yes! done

jessicawyldemoj (approved):

PR #1876, opened 2026-06-29
HIA-1289: Replace (emergency-)contacts and (active-)alerts endpoints …

jessicawyldemoj:
…with filtering on the main endpoints

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsintegrationapi/controllers/v1/person/AlertsController.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsintegrationapi/controllers/v1/person/AlertsController.kt
@@ -51,9 +51,10 @@ class AlertsController(
     @Parameter(description = "The HMPPS ID of the person", example = "A1234AA") @PathVariable hmppsId: String,
     @Parameter(description = "The page number (starting from 1)", schema = Schema(minimum = "1")) @RequestParam(required = false, defaultValue = "1", name = "page") page: Int,
     @Parameter(description = "The maximum number of results for a page", schema = Schema(minimum = "1")) @RequestParam(required = false, defaultValue = "10", name = "perPage") perPage: Int,
+    @Parameter(description = "The status of alerts returned", schema = Schema(minimum = "active")) @RequestParam(required = false, defaultValue = "", name = "status") status: String,
     @RequestAttribute requestContext: RequestContext?,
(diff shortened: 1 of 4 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsintegrationapi/controllers/v1/person/AlertsController.kt
```diff
@@ -51,9 +51,10 @@ class AlertsController(
     @Parameter(description = "The HMPPS ID of the person", example = "A1234AA") @PathVariable hmppsId: String,
     @Parameter(description = "The page number (starting from 1)", schema = Schema(minimum = "1")) @RequestParam(required = false, defaultValue = "1", name = "page") page: Int,
     @Parameter(description = "The maximum number of results for a page", schema = Schema(minimum = "1")) @RequestParam(required = false, defaultValue = "10", name = "perPage") perPage: Int,
+    @Parameter(description = "The status of alerts returned", schema = Schema(minimum = "active")) @RequestParam(required = false, defaultValue = "", name = "status") status: String,
```
pmcphee77: This is an incorrect use of minimum. This is for numeric values. Should be removed and replaced with an example. e.g example = "active" Should this 
`Sets the minimum numeric value for a property. Ignored if the value is an empty string or not a number.`

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsintegrationapi/controllers/v1/person/PrisonerContactsController.kt
```diff
@@ -50,9 +50,10 @@ class PrisonerContactsController(
     @Parameter(description = "The HMPPS ID of the prisoner") @PathVariable hmppsId: String,
     @Parameter(description = "The page number (starting from 1)", schema = Schema(minimum = "1")) @RequestParam(required = false, defaultValue = "1", name = "page") page: Int,
     @Parameter(description = "The maximum number of results for a page", schema = Schema(minimum = "1")) @RequestParam(required = false, defaultValue = "10", name = "perPage") perPage: Int,
+    @Parameter(description = "The type of contacts returned", schema = Schema(minimum = "emergency")) @RequestParam(required = false, defaultValue = "", name = "type") type: String,
```
pmcphee77: Same as above

pmcphee77 (approved):

PR #1883, opened 2026-07-01
HIA-1499: Validate obo username WIP

pmcphee77:
Use the manage users api to verify an obo username


src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsintegrationapi/models/manageUsers/User.kt
```diff
@@ -0,0 +1,6 @@
+package uk.gov.justice.digital.hmpps.hmppsintegrationapi.models.manageUsers
+
+data class User(
```
SteveLeachMoj: Maybe `HmppsAuthUser` to make it clear that these aren't _our_ users (consumers)?

pmcphee77: Done

SteveLeachMoj (approved):

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsintegrationapi/services/ManageUsersService.kt
```diff
@@ -0,0 +1,20 @@
...
+@Service
+class ManageUsersService(
+  private val manageUsersGateway: ManageUsersGateway,
+) {
+  fun usernameExists(
+    username: String,
+    authSources: List<String>,
+  ): Boolean {
+    val usersResponse = manageUsersGateway.findUser(username, authSources)
+    usersResponse.errors.forEach {
```
SteveLeachMoj, line 15: Won't throwing an exception for the first error prevent this from reaching the others?

PR #1901, opened 2026-07-08
HIA-1544: Add a 5 minute cache to the manage users API call

jessicawyldemoj:


src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsintegrationapi/config/CacheConfig.kt
```diff
@@ -81,6 +97,28 @@ class GatewayKeyGenerator : KeyGenerator {
...
+  override fun generate(
+    target: Any,
+    method: Method,
+    vararg params: Any?,
+  ): Any {
+    val keyParts =
+      params.mapNotNull { param ->
+        when (param) {
+          is String -> param // Handles the username
+          is Collection<*> -> param.joinToString("_") // Handles the authSources list
```
pmcphee77: can we change this to be  is `List<*> -> param.filterIsInstance<String>().joinToString("_") `

pmcphee77 (approved):

PR #1907, opened 2026-07-13
HIA-1104: Use paginated person search for probation person search

pmcphee77:
New feature flag `USE_PROBATION_SEARCH_FOR_PERSON_SEARCH`.
When set, probation search will be used for person search


src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsintegrationapi/gateways/PrisonerOffenderSearchGateway.kt
```diff
@@ -144,11 +146,17 @@ class PrisonerOffenderSearchGateway(
   fun attributeSearch(
     request: POSAttributeSearchRequest,
     requestContext: RequestContext? = null,
+    paginatedRequest: PaginatedRequest = PaginatedRequest(),
   ): Response<POSPaginatedPrisoners?> {
+    val uriBuilder = UriComponentsBuilder.fromPath("/attribute-search")
```
jessicawyldemoj: Do we also want to feature flag behind this as well?

src/test/kotlin/uk/gov/justice/digital/hmpps/hmppsintegrationapi/gateways/prisoneroffendersearch/PrisonerOffenderSearchGatewayTest.kt
```diff
@@ -407,7 +407,7 @@ class PrisonerOffenderSearchGatewayTest(
 
         it("returns a prisoner by attributes") {
           prisonerOffenderSearchApiMockServer.stubForPost(
-            "/attribute-search",
+            "/attribute-search?page=0&size=10",
```
jessicawyldemoj: Shoundt we keep test with and without `?page=0&size=10`?

pmcphee77: This is a stub so it requires the `page=0&size=10`

pmcphee77: This gas been reverted as this now behind a feature flag

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsintegrationapi/gateways/PrisonerOffenderSearchGateway.kt
pmcphee77: Good point. Done

jessicawyldemoj (approved):

PR #1912, opened 2026-07-14
HIA-1561: Allow PersonController to pass down OBO username to auth he…

jessicawyldemoj:
…ader


src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsintegrationapi/services/GetPersonService.kt
```diff
@@ -175,6 +175,7 @@ class GetPersonService(
   fun getPersonWithPrisonFilter(
     hmppsId: String,
     filters: ConsumerFilters?,
```
jessicawyldemoj, line 177: This service has a hight blast radius and changing all instances of this service now would be to large of a scope to use only request context. So this service will take in both filters and request context to reduce change radius.

pmcphee77: We have decided to reduce the size of this PR and take this issue out

PR #1918, opened 2026-07-15
HIA-1293 Start converting SANGateway over to RestApiClient

SteveLeachMoj:
Allow the `SANGateway` to use a `RestApiClient` (see https://github.com/ministryofjustice/hmpps-integration-api/pull/1904) instead of a `WebClientWrapper` for the upstream connectivity.

The new client is used if the `restapiclient-for-san-gateway` feature flag is set; for this PR it is only set for Integration Tests. A follow-up PR will enable it in dev, and another one for production. Once we are confident that we won't need to revert we can remove the old `WebClientWrapper` code and the feature flag.

There is also a helper method on the `Response` class to make it easier to create an error response (with a list of `UpstreamApiError` objects) from the result of the `RestApiClient` (which has a list of exceptions).

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsintegrationapi/config/FeatureFlagConfig.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsintegrationapi/config/FeatureFlagConfig.kt
@@ -35,6 +35,7 @@ data class FeatureFlagConfig(
     const val CONTACT_LINKED_PRISONERS_ENDPOINT_ENABLED = "contact-linked-prisoners-endpoint-enabled"
     const val PRISON_TIMELINE_ENDPOINT_ENABLED = "prison-timeline-endpoint-enabled"
     const val USE_PROBATION_SEARCH_FOR_PERSON_SEARCH = "use-probation-search-for-person-search"
+    const val SAN_GATEWAY_USES_RESTAPICLIENT = "san-gateway-uses-restapiclient"
 
     // Events feature flags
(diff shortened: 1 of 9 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsintegrationapi/config/FeatureFlagConfig.kt
```diff
@@ -35,6 +35,7 @@ data class FeatureFlagConfig(
     const val CONTACT_LINKED_PRISONERS_ENDPOINT_ENABLED = "contact-linked-prisoners-endpoint-enabled"
     const val PRISON_TIMELINE_ENDPOINT_ENABLED = "prison-timeline-endpoint-enabled"
     const val USE_PROBATION_SEARCH_FOR_PERSON_SEARCH = "use-probation-search-for-person-search"
+    const val RESTAPICLIENT_FOR_SAN_GATEWAY = "restapiclient-for-san-gateway"
```
jessicawyldemoj, line 38: Are we planning on using feature flags per gateway for this? I would do it as one feature flag for all gateways for the new RestApiClient

SteveLeachMoj: I'm open to guidance on how granular we want to make this, but changing how all our gateways work in one big bang feels a bit risky to me, so would prefer to leave this as-is for now.

pmcphee77 (approved):

PR #1921, opened 2026-07-16
HIA-1570: Alerts endpoint supplied with a CRN is unexpectedly resulti…

jessicawyldemoj:
…ng in a 404

### DIFF
--- a/src/test/kotlin/uk/gov/justice/digital/hmpps/hmppsintegrationapi/services/GetPersonServiceTest.kt
+++ b/src/test/kotlin/uk/gov/justice/digital/hmpps/hmppsintegrationapi/services/GetPersonServiceTest.kt
@@ -911,10 +911,14 @@ internal class GetPersonServiceTest :
           result.data.shouldBe(NomisNumber(nomsNumber))
         }
 
-        it("Returns nomis number when providing hmpps id") {
-          whenever(prisonerOffenderSearchGateway.getPrisonOffender(crnNumber)).thenReturn(Response(data = prisonerWithPrisonId, errors = emptyList()))
+        it("Supervision status check calls NDelius with CRN and Prisons with nomis number for probation only supervision status") {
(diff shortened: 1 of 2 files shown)


src/test/kotlin/uk/gov/justice/digital/hmpps/hmppsintegrationapi/services/GetPersonServiceTest.kt
```diff
@@ -910,6 +910,13 @@ internal class GetPersonServiceTest :
           val result = getPersonService.getNomisNumber(nomsNumber, noneOnlyConsumerFilter)
           result.data.shouldBe(NomisNumber(nomsNumber))
         }
+
+        it("Returns nomis number when providing hmpps id") {
```
pmcphee77: This test isn't actually hitting the problem

The scenario we want is:
1. The consumer has a supervision status of PROBATION Only
2. The person isn't active in Prison.
3. Request is made with a CRN

It should call `prisonerOffenderSearchGateway.getPrisonOffender` with the nomis number, then call deliusGateway.getOffender with the CRN. 

Perhaps replace that with this one. N.B Overall, the deliusGateway.getOffender will be called twice (but its cached so thats ok)

```
  it("Supervision status check calls NDelius with CRN and Prisons with nomis number for probation only supervision status") {
    whenever(prisonerOffenderSearchGateway.getPrisonOffender(nomsNumber)).thenReturn(Response(data = prisonerInactiveOut, errors = emptyList()))
    val result = getPersonService.getNomisNumber(crnNumber, probationOnlyConsumerFilter)
    verify(prisonerOffenderSearchGateway, times(1)).getPrisonOffender(nomsNumber)
    verify(deliusGateway, times(2)).getOffender(crnNumber, null)
    result.data.shouldBe(NomisNumber(nomsNumber))
  }

```

pmcphee77 (approved):

PR #1922, opened 2026-07-16
HIA-1580: Ensure developers are using the MoJ pre-commit hook

jessicawyldemoj:
Here is the MOJ pre commit hooks 
image

### DIFF
--- a/.github/workflows/security-scanning.yml
+++ b/.github/workflows/security-scanning.yml
@@ -12,7 +12,7 @@ jobs:
           fetch-depth: 0
 
       - name: Run Gitleaks
-        uses: gitleaks/gitleaks-action@83d9cd684c87d95d656c1458ef04895a7f1cbd8e # v8.30.1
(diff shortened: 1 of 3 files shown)


.pre-commit-config.yaml
```diff
@@ -1,9 +1,4 @@
 repos:
-  - repo: https://github.com/gitleaks/gitleaks
```
jessicawyldemoj, line 2: MOJ pre hook checks now run gitleaks so we no longer need ours

.github/workflows/security-scanning.yml
jessicawyldemoj, line 1: security-scanning file now no longer uses the pre commit hooks and instead uses and runs a verbose version of Gitleaks and scans only the commits in the pr. This has removed linting done in the security-scanning and let the dedicated pipeline for lint checks to do that instead. We don't need to do lint checks 2 times per pr.

pmcphee77 (approved):

PR #1932, opened 2026-07-21
HIA-1590: Add timezones to return response for prison-timeline times

jessicawyldemoj:


src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsintegrationapi/config/ClockConfiguration.kt
```diff
@@ -12,3 +13,9 @@ class ClockConfiguration {
 }
 
 fun fixedClock(): Clock = Clock.fixed(Clock.systemDefaultZone().instant(), ZoneId.systemDefault())
+
+fun localDateTimeToInstant(dateTime: String): String {
```
SteveLeachMoj: Might be worth calling this `localDateTimeToUkInstant` to be explicit about which timezone it uses.

SteveLeachMoj: Or maybe `ukDateTimeToInstant`

pmcphee77 (approved):

PR #1944, opened 2026-07-27
HIA-1409: Court cases endpoint

pmcphee77:
image

This is the first PR to return the date of first conviction


src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsintegrationapi/gateways/RemandAndSentencingGateway.kt
```diff
@@ -0,0 +1,50 @@
...
+      gitHubRepoUrl = "https://github.com/ministryofjustice/hmpps-remand-and-sentencing-api",
+      slackChannel = "#calculate_release_dates_public_channel",
+    )
+
+  fun getSentencedCourtCases(
+    nomisNumber: String,
+    requestContext: RequestContext? = null,
+  ): Response<RasSentencedCourtCases> {
+    val result =
+      rasRestClient.get(
```
jessicawyldemoj, line 31: Like the san gateway do we want to use both the webClient and the restClient

src/test/kotlin/uk/gov/justice/digital/hmpps/hmppsintegrationapi/mockservers/ApiMockServer.kt
```diff
@@ -43,6 +43,7 @@ class ApiMockServer(
           UpstreamApi.ARNS_INTEGRATION_TEST -> ApiMockServerConfig(4032, "assess-risks-and-needs.json", overrideBindType = false, lenientDateValidation = true)
           UpstreamApi.PROBATION_OFFENDER_SEARCH -> ApiMockServerConfig(4033, "probation-offender-search.json")
           UpstreamApi.MANAGE_USERS -> ApiMockServerConfig(4034, "manage-users.json")
+          UpstreamApi.REMAND_AND_SENTENCING -> ApiMockServerConfig(4036, "remand-and-sentencing.json", true)
```
jessicawyldemoj, line 46: What dose true do here exactly?

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsintegrationapi/controllers/v1/CourtCaseController.kt
```diff
@@ -0,0 +1,57 @@
...
+import org.springframework.web.bind.annotation.RestController
+import uk.gov.justice.digital.hmpps.hmppsintegrationapi.exception.EntityNotFoundException
+import uk.gov.justice.digital.hmpps.hmppsintegrationapi.extensions.RequestContext
+import uk.gov.justice.digital.hmpps.hmppsintegrationapi.models.hmpps.CourtCasesSummary
+import uk.gov.justice.digital.hmpps.hmppsintegrationapi.models.hmpps.DataResponse
+import uk.gov.justice.digital.hmpps.hmppsintegrationapi.models.hmpps.UpstreamApiError
+import uk.gov.justice.digital.hmpps.hmppsintegrationapi.services.CourtCaseService
+import uk.gov.justice.digital.hmpps.hmppsintegrationapi.services.internal.AuditService
+
+@RestController
```
jessicawyldemoj, line 23: Feature flag for new endpoint?

pmcphee77: No, i dont think one is required here as we currently dont have any consumers

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsintegrationapi/gateways/RemandAndSentencingGateway.kt
pmcphee77: No, i dont think so. We want to start moving away from the webclientWrapper

src/test/kotlin/uk/gov/justice/digital/hmpps/hmppsintegrationapi/mockservers/ApiMockServer.kt
pmcphee77: See comment in the BindTypeValidationListener

 ```
In Swagger Core 2.2.x, when processing OpenAPI 3.1 specifications, type field is mapped by Set<String> types
 member (instead of String type) to also support the array data type.
 This means that the foo.type in V31 schema versions is deserialized into a Schema class with populated types (foo.types)
 member (while foo.type remains null).

 This causes the discriminator validation to fail as the type is changed to null (with the following error).
 'discriminator' field 'type' must be defined as a string property

 In order to allow for no change support of 3.1 processing by existing clients in specific limited scenarios,
 users can set system property bind-type=true to have Schema.getType() return the value of the single item of Schema.types
 in OAS 3.1 specifications with a non-array Schema.type.
```

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsintegrationapi/controllers/v1/CourtCaseController.kt
jessicawyldemoj: Can that be mentioned in the ticket then. As its process we add feature flags to new endpoints it should be expressly said we are not doing a endpoint featuer flag.

jessicawyldemoj (approved):

PR #1949, opened 2026-07-29
Hia 1409 court cases remaining fields

pmcphee77:
image

image

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsintegrationapi/controllers/v1/CourtCaseController.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsintegrationapi/controllers/v1/CourtCaseController.kt
@@ -30,7 +30,17 @@ class CourtCaseController(
   @GetMapping
   @Operation(
     summary = "Returns a summary of an offenders court case history.",
-    description = "",
+    description =
+      "Returns information regarding an individuals court case history.<br /> <br />" +
+        "This returns: <br />" +
+        "- The date of their first conviction <br />" +
+        "- The court code of their latest sentenced outcome <br />" +
+        "- The outcome of their last court appearance including outcome name and outcome code. e.g Imprisonment (SENTENCING), Appeal (APPEAL) or Remand (Remand) <br /><br />" +
+        "The following outcome type codes are supported: <br />" +
+        "- SENTENCING </br>" +
+        "- APPEAL </br>" +
+        "- REMAND </br>" +
+        "- OTHER  </br>",
     responses = [
       ApiResponse(responseCode = "200", useReturnTypeSchema = true, description = "Successfully found court cases summary."),
       ApiResponse(responseCode = "404", content = [Content(schema = Schema(ref = "#/components/schemas/PersonNotFound"))]),
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsintegrationapi/controllers/v1/CourtCaseController.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsintegrationapi/controllers/v1/CourtCaseController.kt
@@ -30,17 +30,6 @@ class CourtCaseController(
   @GetMapping
   @Operation(
     summary = "Returns a summary of an offenders court case history.",
-    description =
-      "Returns information regarding an individuals court case history.<br /> <br />" +
-        "This returns: <br />" +
-        "- The date of their first conviction <br />" +
-        "- The court code of their latest sentenced outcome <br />" +
-        "- The outcome of their last court appearance including outcome name and outcome code. e.g Imprisonment (SENTENCING), Appeal (APPEAL) or Remand (Remand) <br /><br />" +
-        "The following outcome type codes are supported: <br />" +
-        "- SENTENCING </br>" +
-        "- APPEAL </br>" +
-        "- REMAND </br>" +
-        "- OTHER  </br>",
     responses = [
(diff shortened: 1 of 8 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsintegrationapi/models/remandAndSentencing/SentencedCourtCase.kt
```diff
@@ -37,3 +38,14 @@ data class RasCharge(
 data class RasSentence(
   val convictionDate: LocalDate? = null,
 )
+
+enum class CourtOutComeType {
+  SENTENCING,
+  APPEAL,
+  OTHER,
```
SteveLeachMoj, line 46: Should this include REMAND?

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsintegrationapi/controllers/v1/CourtCaseController.kt
```diff
@@ -30,7 +30,17 @@ class CourtCaseController(
...
   @Operation(
     summary = "Returns a summary of an offenders court case history.",
-    description = "",
+    description =
+      "Returns information regarding an individuals court case history.<br /> <br />" +
+        "This returns: <br />" +
+        "- The date of their first conviction <br />" +
+        "- The court code of their latest sentenced outcome <br />" +
+        "- The outcome of their last court appearance including outcome name and outcome code. e.g Imprisonment (SENTENCING), Appeal (APPEAL) or Remand (Remand) <br /><br />" +
+        "The following outcome type codes are supported: <br />" +
```
SteveLeachMoj: Is this just duplicating the content of the schema definition?

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsintegrationapi/models/remandAndSentencing/SentencedCourtCase.kt
```diff
@@ -37,3 +38,14 @@ data class RasCharge(
...
+  ;
+
+  companion object {
+    fun from(type: String?): CourtOutComeType = CourtOutComeType.entries.firstOrNull { it.name == type } ?: OTHER
```
SteveLeachMoj, line 50: Can @JsonEnumDefaultValue handle this?

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsintegrationapi/services/CourtCaseService.kt
```diff
@@ -34,6 +36,16 @@ class CourtCaseService(
...
-    return Response(CourtCasesSummary(dateOfFirstConviction = dateOfFirstConviction))
+    val allCourtAppearances =
+      sentencedCourtCases.data.courtCases
+        .flatMap { case ->
+          (case.appearances + listOf(case.latestAppearance))
+        }.sortedByDescending { it?.appearanceDate }
+
+    val courtOutcome = allCourtAppearances.firstOrNull()?.outcome?.let { CourtOutcome(CourtOutComeType.from(it.outcomeType), it.outcomeName) }
+
+    val courtCode = allCourtAppearances.firstOrNull { it?.outcome?.outcomeType == CourtOutComeType.SENTENCING.name }?.courtCode
```
SteveLeachMoj, line 47: Does this result in null if there was a non-SENTENCING appearance (such as an appeal) after the sentencing hearing? Can that happen, and is it what we want to return in that scenario?

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsintegrationapi/models/remandAndSentencing/SentencedCourtCase.kt
pmcphee77: This isnt really something id like to use here. This means we would need to add the enum to the  upstreams model. Also READ_UNKNOWN_ENUM_VALUES_USING_DEFAULT_VALUE would need to be set on the object mapper (it isnt by default). I would prefer to do it this way without using annotations. Gives me more control and is more testable

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsintegrationapi/services/CourtCaseService.kt
pmcphee77: no, this will NOT return null if there was a non-sentencing appearance after the sentencing hearing. It will pick the first (latest) sentencing appearance. If there is no sentencing appearances only then will it return NULL

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsintegrationapi/models/remandAndSentencing/SentencedCourtCase.kt
pmcphee77: Yes - ive added this

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsintegrationapi/controllers/v1/CourtCaseController.kt
pmcphee77: Removed the description

SteveLeachMoj (approved):

PR #1953, opened 2026-07-30
HIA-1472: Connect Assessment Summary endpoint to ARNS API

jessicawyldemoj:


src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsintegrationapi/models/assessRisksAndNeeds/ArnAssessmentSummary.kt
```diff
@@ -4,20 +4,30 @@ import uk.gov.justice.digital.hmpps.hmppsintegrationapi.models.hmpps.AssessmentS
 import java.time.LocalDateTime
 
 data class ArnAssessmentSummary(
+  val assessments: List<ArnAssessmentSummaryItem>? = null,
+) {
+  fun toAssessmentSummary(): AssessmentSummary {
+    val latestAssessment =
+      this.assessments
+        ?.filter { it.dateCompleted != null }
+        ?.maxByOrNull { it.dateCompleted!! }
```
pmcphee77: This would throw a null pointer if no assessment summaries with a completed date are returned.

```
    val latestAssessment = this.assessments?.filter { it.dateCompleted != null }?.sortedByDescending { it.dateCompleted }?.firstOrNull()
      ?: throw EntityNotFoundException("No assessment summary found")
```
This will then return a 404 if no assessment can be found and we can be sure that this method will always return a single assessment

scripts/K6/smoke-tests.js
```diff
@@ -121,6 +121,7 @@ const get_endpoints = [
   `/v1/activities/schedule/${scheduleId}/suitability-criteria`,
   `/v1/persons/${hmppsId}/court-cases`,
   `/v1/hmpps/reference-data/courts/${courtId}`
+  `/v1/persons/${hmppsId}/assessment-summary`
```
SteveLeachMoj: This list is the "legacy" part of the smoke test script, for endpoints that we haven't yet implemented structured tests for. See `verify_risk_endpoints`. Ideally we would test that we actually get assessment data from upstream.

pmcphee77 (approved):

PR #1955, opened 2026-08-03
Hia 1635 redactions refactor

SteveLeachMoj:
The redactions package and sub-packages are currently a bit disorganised and could do with tidying up.

- One of the redaction policies is in the dsl package for some reason
- Other than that there is only a single class in the dsl package
- The RedactionContext utility class is mixed in with the redaction policies

We should create policies sub-package and move all the policies into that, then move the DSL code into the top redactions package and remove the dsl sub-package.

While doing this work we found that the `ObjectMapper` instance from the `redactions` package was being used all over the code, so this has also been refactored out into a more sensible location.


src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsintegrationapi/config/JsonConfig.kt
jessicawyldemoj, line 1: I think Json.kt is a bit too vague of a file name to know this is where the defaultObjectMapper is found. 

As this file only has the defaultObjectMapper would a better name be for the file DefaultObjectMapper.kt?

SteveLeachMoj: We could do, but that would mean that this would forever be a one-variable file. I was thinking this could be just the place we put all JSON handling code, with the `defaultObjectMapper` just being the first thing in there.

SteveLeachMoj: Moving the file to `config/JsonConfig.kt`

jessicawyldemoj (approved):

PR #1958, opened 2026-08-04
HIA-1561: Allow SANController to pass down OBO username to auth header

jessicawyldemoj:


src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsintegrationapi/gateways/SANGateway.kt
```diff
@@ -35,16 +36,19 @@ class SANGateway(
   @Autowired
   lateinit var hmppsAuthGateway: HmppsAuthGateway
 
-  fun getPlanCreationSchedules(prisonerNumber: String): Response<PlanCreationSchedules> {
+  fun getPlanCreationSchedules(
+    prisonerNumber: String,
+    requestContext: RequestContext?,
+  ): Response<PlanCreationSchedules> {
     if (useRestApiClient()) {
```
SteveLeachMoj: Note the following from the acceptance criteria in the ticket...

"The feature flag check in SANGateway should be modified to use the feature flags from the request context, and not from an injected Spring Bean."

zaheeriqbalMoJ (approved):

PR #1966, opened 2026-08-07
HIA-1420 Consolidate certificate handling code in CertificateService

SteveLeachMoj:
The code for processing client certificate details was split somewhat arbitrarily across the `ConsumerFilter` and `AuthorisationService`.

In `ConsumerFilter`...
* `extractCertificateSerialNumber`
* `extractConsumerName`
* `certificateRevoked`

In `AuthorisationService`...
* `processCertificateExpiryDate`
* `checkExpiryDate`
* `expiryWarningMessage`

This PR consolidates all of this into a new `CertificateService`.

The `CertificateService` will also be a good place to put any code for renewing certificates if we implement an endpoint for client certificate renewal (see https://dsdmoj.atlassian.net/browse/HIA-1182).


src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsintegrationapi/services/CertificateService.kt
```diff
@@ -0,0 +1,178 @@
...
+import java.util.Locale
+import kotlin.math.ceil
+
+@Component
+class CertificateService(
+  private val authorisationConfig: AuthorisationConfig,
+  private val telemetryService: TelemetryService,
+  private val clock: Clock = fixedClock(),
+) {
+  fun validateCertificate(
```
pmcphee77: The name of this function feels a bit wrong:
Should it be something like `getCertificateInfo`

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsintegrationapi/services/CertificateService.kt
```diff
@@ -0,0 +1,178 @@
...
+    }
+    return false
+  }
+
+  /**
+   * Converts the certificate serial number sent in the header into hex format
+   * e.g 9572494320151578633330348943480876283449388176
+   * becomes 01:7B:EB:77:06:DB:11:F5:2E:B6:F7:37:7B:A9:E0:E4:84:C5:2C:A3
+   */
+  fun extractCertificateSerialNumber(serialNumber: String?): String? =
```
pmcphee77: Would it be better to call this function `toHexFormat` ?

